PR #8, opened 2021-06-19
Support prototyping strings, variables, operators and expressions

wixette:
Example commands:

1) Generate a simple string:

node . -b string -o out.svg -c 'Hello, World!'

b

2) Generate an expression that contains a number of number values, operators and variables:

node . -b expression -o out.svg -c '130,÷,num,-,(,factor,^,7e4,+,delta,),×,-3.1415926,>=,7,and,1,==,2'

a

### DIFF
--- a/utils/block_proto/app.js
+++ b/utils/block_proto/app.js
@@ -32,17 +32,20 @@ import {getBlockList, render} from './blocks.js';
 const argv = yargs(hideBin(process.argv)).options({
   'block': {
     alias: 'b',
+    type: 'string',
     describe: 'The kind of the main block to output.',
     demandOption: true,
     choices: getBlockList(),
   },
   'config': {
     alias: 'c',
+    type: 'string',
     describe: 'Optional config string for rendering blocks.',
     demandOption: false,
   },
   'output': {
     alias: 'o',
+    type: 'string',
     describe: 'The output SVG file path.',
     demandOption: false,
   },
--- a/utils/block_proto/app.js
+++ b/utils/block_proto/app.js
@@ -19,11 +19,16 @@
  * @fileoverview The main entry of the block prototype maker.
  */
 
+// @ts-check
+
 import fs from 'fs';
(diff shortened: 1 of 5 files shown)


utils/block_proto/app.js
```diff
@@ -19,31 +19,39 @@
  * @fileoverview The main entry of the block prototype maker.
  */
 
+// @ts-check
+
 import fs from 'fs';
```
codingpotato, line 24: There is a jshint warning "'import' is only available in ES6 (use 'esversion: 6'). (W119)jshint(W119)" in my vscode environment. Same warning happens for 'const', 'of', etc. Need add config to eliminate this warning?

utils/block_proto/blocks.js
```diff
@@ -71,35 +120,167 @@ export function render(block, config) {
...
+
+  draw.rect(shapeWidth, shapeHeight)
+      .fill(blockDef.background)
+      .radius(GLOBAL_DEFS.rectRadius)
+      .move(shapeOffset.x, shapeOffset.y);
+
+  renderCenterText(draw, valueString, shapeWidth, shapeHeight, blockDef.color,
+      shapeOffset);
+
+  if (blockDef.delimiter) {
```
codingpotato, line 185: Is it possible to add delimiters at the beginning and end of the valueString, then render the whole string with delimiters, to avoid the calculation of the delimiters' position?

wixette: Delimiters have a separate text color different from the main text.

utils/block_proto/app.js
wixette: For VS Code + JS development, jslint is kind of obsolete. E.g., there is a config line in the package.json:

"type": "module",

jslint should have taken this config line and turn ES6module mode on by default.

Use ESLint instead. ESLint is the recommended lint tool integrated with VS Code.

codingpotato (approved):

PR #10, opened 2021-06-22
Add byte code design

codingpotato:
Add description and layout of byte code, and some simple instructions for calculation support.


design/bytecode.md
```diff
@@ -0,0 +1,106 @@
+# Byte code
+
```
wixette, line 2: I think it would be great to have a short "background" or "introduction" section before everything else. For example, it is super helpful to clarify in this short section that:

- SeedInterpreter will be a register-based virtual machine.
- The design of the instructions is a derived version of Lua byte code instructions.

A reference link to [The Implementation of Lua 5.0](https://www.lua.org/doc/jucs05.pdf) or [A No-Frills Introduction to Lua 5.1 VM Instructions](https://talk.pokitto.com/uploads/default/original/2X/7/716c67a0b5b1636cbc4dc1fec232ca2536cb74d1.pdf) may also help.

design/bytecode.md
```diff
@@ -0,0 +1,106 @@
...
+## Instruction Notation
+
+Following notations are used to explain the behavior of the instructions.
+
+- PC: Program Counter
+- R[A]: Register at index A
+- R[B]: Register at index B
+- R[C]: Register at index C
+- K[N]: Element N in the constant list
+- RK[C]: Register at index C (when K = 0) or constant at index C (K = 1)
```
wixette, line 48: It is confusing to have both the constant flag K and the RK[C] notation in the same context.

As I see in Lua's VM design, notation K is only used for "element n in the constant list". Thus, RK[C] is easy to understand in Lua's context.

In this design, you use K as a flag bit to indicate whether the operand C is constant or not. Given this context, K[Bx] looks still okay but "RK[C] ... when K = 0" looks a bit weird since K appears twice in the same line, with different meanings.

How about using another notation for the constant flag? E.g., KF, for "constant flag"?

design/bytecode.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+Copy the value of the register R[B] into register R[A]. If R[B] holds a object,
+the reference to that object is copied. Move is often used for moving values
+into place for the next operation.
+
+## LoadI Instruction
+
+```text
+LoadI A sBx         // R[A] := sBx
```
wixette, line 63: Is this an equivalent but a quicker (immediate value) version of LoadK?

design/bytecode.md
```diff
@@ -0,0 +1,106 @@
...
+  (Max = 2 ^ 16). The actual value is the unsigned value (Bx) minus Max, so that
+  -Max is represented by 0, and Max - 1 is represented by 2 \* Max - 1.
+
+The layout of the instructions is illustrated as below:
+
+```text
+
+     3 3 2 2 2 2 2 2 2 2 2 2 1 1 1 1 1 1 1 1 1 1 0 0 0 0 0 0 0 0 0 0
+     1 0 9 8 7 6 5 4 3 2 1 0 9 8 7 6 5 4 3 2 1 0 9 8 7 6 5 4 3 2 1 0
+    |K|      C (8)    |      B (8)    |     A (8)     |   Op (7)    |
```
wixette, line 24: Per the below design of AddI, I guess a better idea is to have a separate line for sC, e.g.:

| sC(9) | B(8) | A (8) | Op (7) |

So that the flag K in this graph won't need to have two different meanings (a constant flag, and, a sign flag).

design/bytecode.md
```diff
@@ -0,0 +1,106 @@
...
+
+Load the constant of K[Bx] into the register R[A].
+
+## AddI Instruction
+
+```text
+AddI A B sC         // R[A] := R[B] + sC
+```
+
+Add the value of the register R[B] and the constant number sC (positive if K
```
wixette, line 83: As I pointed out in the above comments, if there is a layout design like

| sC(9) | B(8) | A (8) | Op (7) |

K will not need to represent both a constant flag and a sign flag, which is really confusing here.

design/bytecode.md
```diff
@@ -0,0 +1,106 @@
...
+
+```text
+Add A B C           // R[A] := R[B] + R[C]
+Sub A B C           // R[A] := R[B] - R[C]
+Mul A B C           // R[A] := R[B] * R[C]
+Div A B C           // R[A] := R[B] / R[C]
+```
+
+Perform the binary operators between the register R[B] and the constant K[C] or
+register R[B]. The result of the operation is placed into R[A]. These
```
wixette, line 105: R[B] -> R[C]

design/bytecode.md
```diff
@@ -0,0 +1,106 @@
...
+instructions are in the classic three address style.
```
wixette, line 106: Lua byte code allows the operand B to be either a register or a constant - the RK[B] notation. What are your considerations to only allow the third operand to be an index to a constant value? What is the trade-off here?

As for "Constant Folding", E.g., see page 25 of  [A No-Frills Introduction to Lua 5.1 VM Instructions ](https://talk.pokitto.com/uploads/default/original/2X/7/716c67a0b5b1636cbc4dc1fec232ca2536cb74d1.pdf), there might be some cases where constant folding is not performed:

local a = 1 / 0; local b = 1 + "1"
[1] div 0 256 257 ; 1 0
[2] add 1 256 258 ; 1 "1"

design/bytecode.md
```diff
@@ -0,0 +1,106 @@
+# Byte code
+
+The Byte code of Seed is generated from the SeedIR code, and executed by the
```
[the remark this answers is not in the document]
codingpotato: Will fix it

design/bytecode.md
codingpotato: Good suggestion. Will add it.

design/bytecode.md
codingpotato: The reason of the confusion might be the difference between Lua 5.4 change, which is not documented. Current design follows the concept of Lua 5.4 design.

The instruction layout of Lua 5.0-5.3 is:
| op(6) | A(8) | B(9) | C(9) |

So both B and C support RK() operation to select register or constant based on the 9th bit. The binary operation should be:
ADD A B C         // R[A] := RK[B] + RK[C]

Lua 5.4 changes the layout to:
| op(7) | A(8) |K| B(8) | C(8) |

I think the intention is to double supported opcode numbers to support both ADDK and ADD, so that one condition check for RK() operation can be avoid. The binary operation should be:
ADDK A B C      // R[A] := R[B] + K[C]              K[B] + K[C] should be optimized into compile time calculation
ADD A B C        // R[A] := R[B] + R[C]

There are some instruction in Lua 5.4 uses RK() operation for array element and class self access. I'm not sure if RK() operation can be avoid now.

Maybe we can have several options:
1. Follow Lua 3.5's layout, and ignore Lua 5.4's optimization. It might not be a significant improvement.
2. Follow Lua 5.4's layout, but combine C and K to a 9 bits C for RK(C) support. ADDI is a very small optimization, may not needed.

design/bytecode.md
codingpotato: It's just a optimization for LoadK to reduce the size of constant list. And the speed will be a little quicker.

design/bytecode.md
codingpotato: AddI is also a optimization for adding and subtracting small number. It can be replaced by AddK. Is it better to not add it in the first version?

design/bytecode.md
codingpotato: See above discussion.

For the "Constant Folding". I need check how Lua 5.4 handle it. I guess one alternative method might be LoadK into a register first, and then use this register as the first operand.

design/bytecode.md
codingpotato: Will fix it

design/bytecode.md
codingpotato: I found another instruction in Lua 5.4:

OP_EQI,/*	A sB k	if ((R[A] == sB) ~= k) then pc++		*/

So sB and K are used at the same time. Maybe K is not used as a sign flag. and the range of sB and sC should be in [-128, 128), instead of [-256, 256).

wixette:
Sorry. I never noticed that there are so many significant changes in the instruction set design between Lua VM v5.3 and v5.4. If our bytecode design is going to follow Lua's VM somehow, we need a sharp decision on the reference Lua VM version. It might not be a good idea to mix the design decisions of the two versions.

Let's have a chat before moving forward on this pr.

design/seed_vm.md
```diff
@@ -0,0 +1,190 @@
...
+iABC    | Opcode: 6 |     A: 8      |      B: 9       |      C: 9       |
+        -----------------------------------------------------------------
+iABx    | Opcode: 6 |     A: 8      |              Bx: 18               |
+        -----------------------------------------------------------------
+iAsBx   | Opcode: 6 |     A: 8      |             sBx: 18               |
+        -----------------------------------------------------------------
+```
+
+There are at most 64 opcodes for SeedLang byte code, because the opcode occupies
+the lower 6 bits of instruction.
```
wixette, line 60: of instruction -> of the instruction

design/seed_vm.md
```diff
@@ -0,0 +1,190 @@
...
+
+Instruction fields are encoded as simple unsigned integer values, except for
+sBx. Field sBx represents negative numbers, but it doesn't use 2s complement.
```
wixette, line 63: Field sBx -> The field sBx

2s complement -> "two's complement" or "2's complement"

design/seed_vm.md
```diff
@@ -0,0 +1,190 @@
...
+Instead, it has a bias equal to half the maximum integer that can be represented
+by its unsigned counterpart, Bx. For a field size of 18 bits, Bx can hold a
+maximum unsigned integer value of 262143, and so the bias is 131071 (calculated
+as 262143 >> 1). A signed value of x will be encoded as (x + 131071).
```
wixette, line 67: It would be better to also note that the range of sBx is [-131071, 131072]

design/seed_vm.md
```diff
@@ -0,0 +1,190 @@
...
+No-Frills Introduction to Lua 5.1 VM
+Instructions](http://underpop.free.fr/l/lua/docs/a-no-frills-introduction-to-lua-5.1-vm-instructions.pdf).
+
+## SeedLang Value Types
+
+SeedLang is a dynamically typed programming language. Types are attached to
+values rather then variables. There are 9 types in SeedLang: Nil, Boolean,
+Number, String, List, Table, Function, Module and SeedObject:
+
+- Nil: Nil value, only one value nil
```
wixette, line 26: I know it is a bit annoying but also worth doing this: add markdown code highlighting notation <code>\`...\`</code> to all identifiers, keywords, register names, and instruction names, so that the syntax highlighting will be way better than the default text style.

For example:

- `Nil`: Nil value, only one value `nil`.

design/seed_vm.md
```diff
@@ -0,0 +1,190 @@
...
+Instead, it has a bias equal to half the maximum integer that can be represented
+by its unsigned counterpart, Bx. For a field size of 18 bits, Bx can hold a
+maximum unsigned integer value of 262143, and so the bias is 131071 (calculated
+as 262143 >> 1). A signed value of x will be encoded as (x + 131071).
+
+## Instruction Notation
+
+| Notation         | Description                                      |
+| ---------------- | ------------------------------------------------ |
+| PC               | Program Counter                                  |
```
wixette, line 73: As I mentioned above, it will be way better to add the code highlighting notation to "PC", "R(A)", "A", "B", "C", etc. inside this table.

design/seed_vm.md
```diff
@@ -0,0 +1,190 @@
...
+- MSB == 1: (B - 256) or (C - 256) is the index of the constant in the constant
+  list
+
+## Instruction Set
+
+All the SeedLang instructions are listed as follows:
+
+| Opcode | Name     | Description                                         |
+| :----: | -------- | --------------------------------------------------- |
+|   0    | MOVE     | Copy a value between registers                      |
```
wixette, line 93: As I mentioned above, it will be way better to add the code highlighting notation to the instruction names inside this table.

design/seed_vm.md
```diff
@@ -0,0 +1,190 @@
...
+|   19   | LE       | Less than or equal to test                          |
+|   20   | TEST     | Boolean test, with conditional jump                 |
+|   21   | TESTSET  | Boolean test, with conditional jump and assignment  |
+|   22   | CALL     | Call a function                                     |
+|   23   | RETURN   | Return from a function call                         |
+
+### Move and Load Constant
+
+```shell
+MOVE A B                    // R(A) := R(B)
```
wixette, line 121: Since you have annotated this markdown code block as a "shell" code, it's better to start the comments inside the code with a real shell programming notation, i.e., "#", instead of "//". By using "#", the rendered markdown doc will look pretty when the comments are highlighted.

Same for the following instructions.

design/seed_vm.md
```diff
@@ -0,0 +1,190 @@
...
+Number, String, List, Table, Function, Module and SeedObject:
+
+- Nil: Nil value, only one value nil
+- Boolean: Boolean value, true or false
+- Number: Double precision floating point number
+- String: A sequential collection of characters to represent a text
+- List: A list of values that can be access by index
+- Table: A dictionary to hold key and value pairs
+- Function: Function object which includes module level function, member
+  function or build-in function
```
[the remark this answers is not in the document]
codingpotato: Agree. Build-in function might be a module-level function as well.

design/seed_vm.md
```diff
@@ -0,0 +1,190 @@
...
+
+## SeedLang Value Types
+
+SeedLang is a dynamically typed programming language. Types are attached to
+values rather then variables. There are 9 types in SeedLang: Nil, Boolean,
+Number, String, List, Table, Function, Module and SeedObject:
+
+- Nil: Nil value, only one value nil
+- Boolean: Boolean value, true or false
+- Number: Double precision floating point number
```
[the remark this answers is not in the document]
codingpotato: Boxing and unboxing is used in C# for transferring numbers from the stack to the heap. We don't need such kind of operation in SeedLang.
My first feeling is that we may not need the boxing and unboxing operation in current phase, because SeedLang is a dynamic programming language. All the values are objects from user's point of view. Compiler will translate the function call (3).toString() to a build-in function sys.__toString__(3). The first 3 is the object 3 in SeedAST, and the second 3 is the SeedVM value 3 with Number type. Low level implement will not impact the hight level abstract.
So I don't want to introduce the complexity of Number and number now. But I do add a description to clarify it:
```
All the SeedLang values are objects from SeedLang's perspective. They all
support properties and functions like `name`, `type`, `doc`, `toString()`, etc.
But in SeedVm, Nil, Boolean and Number are implemented by primitive number
types, other types are implemented by objects of the host language.
```

design/seed_vm.md
```diff
@@ -0,0 +1,190 @@
...
+- String: A sequential collection of characters to represent a text
+- List: A list of values that can be access by index
+- Table: A dictionary to hold key and value pairs
+- Function: Function object which includes module level function, member
+  function or build-in function
+- Module: Module object which can hold functions and objects in it
+- SeedObject: The base type of all the customer objects
```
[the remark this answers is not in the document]
codingpotato: I decide to remove SeedObject. My intention of adding SeedObject is to implement following SeedPython code:
```python
obj = SeedObject({
    'x': 10,
    'y': 20,
    'add': m_add,
    'inner_obj': {
        'x': 100,
        'y': 200,
        'add': m_add,
    },
})
```
But it's easy for compiler to translate this code into a Table object. As we only support object now. the name of obj is enough to identify the object.
We may need a SeedObject base class for all the SeedAST or SeedVM values. But this might be a host language class and implement detail, no need to state here.

design/seed_vm.md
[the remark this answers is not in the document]
wixette: OK.

design/seed_vm.md
```diff
@@ -0,0 +1,198 @@
...
+values rather than variables. There are 8 value types in SeedLang: `Nil`,
+`Boolean`, `Number`, `String`, `List`, `Table`, `Function` and `Module`.
+
+Values of all types are the first-class values. They can be stored in global
+variables, local variables, object properties, list and table fields. They can
+be passed as arguments to functions, and returned from functions as well.
+
+All the SeedLang values are objects from SeedLang's perspective. They all
+support properties and functions like `name`, `type`, `doc`, `toString()`, etc.
+But in SeedVm, `Nil`, `Boolean` and `Number` are implemented by primitive number
```
wixette, line 32: are implemented by -> are implemented with

design/seed_vm.md
```diff
@@ -0,0 +1,198 @@
...
+types, other types are implemented by objects of the host language.
```
wixette, line 33: are implemented by -> are implemented with

design/seed_vm.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Move and Load Constant
+
+```shell
+MOVE A B                    # R(A) := R(B)
+LOADBOOL A B C              # R(A) := (Bool)B; if C then PC++
+LOADK A Bx                  # R(A) := Kst(Bx)
+```
+
+`MOVE` instruction is not needed for arithmetic expression. All arithmetic
```
wixette, line 133: arithmetic expression -> arithmetic expressions

wixette (approved):

Issue #17, opened 2021-07-10
Define project structure for C# libraries


codingpotato:

codingpotato:
Implemented in PR #20

PR #20, opened 2021-07-12
Create csharp solution and projects

codingpotato:
Implement issue #17

- Name of classes and methods are just placeholders, will be changed in the future
- Editorconfig cannot handle semicolon in the file header. So unset file_header_template in the .editorconfig now


csharp/src/SeedLang.Ast/SeedLang.Ast.csproj
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
wixette, line 4: Current Unity LTS releases do not support .Net 5.0. Which framework should we use to build C# assemblies?

Probably switching to .Net 4.x then waiting for Unity's coming upgrades?

I'd be great to test this empty DLL against a Unity test project to make sure the project settings are compatible with the Unity environment.

csharp/src/SeedLang.Block/SeedLang.Block.csproj
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
wixette, line 4: Framework version: 4.x or 5.0?

csharp/src/SeedLang.VM/SeedLang.VM.csproj
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
wixette, line 4: Framework version: 4.x or 5.0?

csharp/src/SeedLang.X/SeedLang.X.csproj
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
wixette, line 4: Framework version: 4.x or 5.0?

csharp/src/SeedLang/SeedLang.csproj
```diff
@@ -0,0 +1,8 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5.0</TargetFramework>
```
wixette, line 5: Framework version: 4.x or 5.0?

csharp/src/SeedLang.Ast/SeedLang.Ast.csproj
codingpotato: After a quick searching, I found the dotnet framework 4.x version is not cross-platform version. It can only run on Windows. Is it possible to use this target on macOS?

wixette: Let's do some quick experiments to make sure which .net version is the best fit here.

wixette: Here is my findings:

1) .NET project can be configured to build multiple targets for more than one frameworks, e.g.:

```
    <TargetFrameworks>net5.0;netstandard2.0;netstandard2.1;netcoreapp3.1</TargetFrameworks>
```

To list all the available frameworks that the current platform supports:

`dotnet new classlib --help`

2) Unity can load and run .Net assemblies targeted to different frameworks. E.g., Unity can load and run net5.0 DLL correctly per my test. Even when I introduced some new [C# 9](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9) feature into the source code of the net5.0 DLL, Unity worked with that DLL too.

3) I used RuntimeInformation.FrameworkDescription to print out the framework version in Unity runtime. The result got from MacOS, Unity 2020.3.11 is:

`Mono 5.11.0 ((HEAD/9283a23f456)`

Overall, my guess is, Unity is able to load and run net5.0 DLLs but is only limited to the ability (e.g., the supported APIs) of the hosting framework. In MacOS, the limit is set by Mono 5.11 - we cannot use the APIs that Mono 5.11 doesn't support.

Per the doc of [.Net standard](https://docs.microsoft.com/en-us/dotnet/standard/net-standard), Mono 5.x is at the level of netstandard2.0. It's said that Unity will support netstandard2.1 (and, probably .Net 5.0 as well) in the coming release of 2021.2.x.

I think we can build our code to both net5.0 and netstandard2.0, so that we have the flexibility to choose a DLL version when we embed it to a Unity project. This solution limits us to using netstandard2.0 features for now. Going forward, we can remove such restriction once Unity officially supports .Net 5.0. For example, config our C# projects as below:

```
    <TargetFrameworks>net5.0;netstandard2.0</TargetFrameworks>
```

Per another [.Net 5.0 doc](https://docs.microsoft.com/en-us/dotnet/core/dotnet-five), here is something interesting:

> For .NET 5.0 apps and libraries, the net5.0 Target Framework Moniker (TFM) combines and replaces the netcoreapp and netstandard TFMs. However, if you plan to share code between .NET Framework, .NET Core, and .NET 5 workloads, you can do so by specifying netstandard2.0 as your TFM. For more information, see .NET Standard.

Your thoughts?

wixette: More references:

https://visualstudiomagazine.com/articles/2021/04/22/unity-plans.aspx

> Complete C# 9 support: "We'd like to have covariant return type support in Mono and IL2CPP for Unity 2021.2. This will bring full C# 9 support to Unity users as well."

> .NET 6 support: Peterson described this as "still a bit fuzzy," with the company hoping to provide a preview around the time the landmark, unifying, milestone .NET 6 ships in November as a long-term support (LTS) release, meaning that "Unity will likely skip .NET 5 entirely." 

https://github.com/mob-sakai/CSharpCompilerSettingsForUnity

> This package changes the C# compiler (csc) used on your Unity project, to support C# 8.0.

.gitignore
```diff
@@ -142,6 +142,392 @@ dist
...
+*.mm.*
+AutoTest.Net/
+
+# Web workbench (sass)
+.sass-cache/
+
+# Installshield output folder
+[Ee]xpress/
+
+# DocProject is a documentation generator add-in
```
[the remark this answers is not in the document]
codingpotato: Done

csharp/src/SeedLang.Ast/SeedLang.Ast.csproj
codingpotato: Add "netstandard2.0" is the best choice now. And we might not need C# 9, the features of C# 8 is enough for us now. With this change, we can do development in macos environment. Do we need a Windows docker image running for Github action to build the release images?

csharp/src/SeedLang/SeedLang.csproj
codingpotato: Done

csharp/src/SeedLang.Block/SeedLang.Block.csproj
codingpotato: Done

csharp/src/SeedLang.VM/SeedLang.VM.csproj
codingpotato: Done

csharp/src/SeedLang.X/SeedLang.X.csproj
codingpotato: Done

csharp/src/SeedLang.Ast/SeedLang.Ast.csproj
wixette: For continuous integration, GitHub has a Windows VM for us to run build and unit test tasks.

wixette (approved):

PR #22, opened 2021-07-13
Add the test task for VS Code

wixette:


.vscode/tasks.json
```diff
@@ -37,6 +37,19 @@
                 "/consoleloggerparameters:NoSummary"
             ],
             "problemMatcher": "$msCompile"
+        },
```
codingpotato, line 40: Different indent from my previous commit. Shall we add any config to enfore the format of vscode settings?

codingpotato (changes requested):

.vscode/tasks.json
wixette: Let's follow your indentions for now.

codingpotato (approved):

PR #24, opened 2021-07-17
Generate ANTLR4 parser

codingpotato:
- Fix license of SeedPython grammar file, the content of all the grammars will be refined in the future
- Change name of SeedAST.asdl
- Add Bash script to generate ANTLR4 parser
- Add step in Github csharp workflow to generate ANTLR4 parser

### DIFF
--- a/grammars/SeedPython.g4
+++ b/grammars/SeedPython.g4
@@ -1,9 +1,24 @@
 /*
+ * Copyright 2021 The Aha001 Team.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * 
+ * This is a derived work of https://github.com/bkiers/Python3-parser.
+ * Here is the original license notice:
+ *
  * The MIT License (MIT)
  *
- * Copyright (c) 2014 Bart Kiers
- * Copyright (c) 2019 Robert Einhorn
- * Copyright (c) 2021 The Aha001 Team.
+ * Copyright (c) 2014 by Bart Kiers
  *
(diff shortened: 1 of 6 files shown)


grammars/SeedPython.g4
```diff
@@ -44,27 +102,44 @@ SUB: '-';
...
+OPEN_BRACK: '[';
+CLOSE_BRACK: ']';
+OPEN_BRACE: '{';
+CLOSE_BRACE: '}';
+
+SKIP_: ( SPACES | COMMENT | LINE_JOINING) -> skip;
+
+UNKNOWN_CHAR: .;
+
+/*
```
wixette, line 128: Delete the trailing space.

grammars/SeedPython.g4
```diff
@@ -44,27 +102,44 @@ SUB: '-';
...
+CLOSE_BRACK: ']';
+OPEN_BRACE: '{';
+CLOSE_BRACE: '}';
+
+SKIP_: ( SPACES | COMMENT | LINE_JOINING) -> skip;
+
+UNKNOWN_CHAR: .;
+
+/* 
+ * Fragments
```
wixette, line 129: Delete the trailing space.

grammars/SeedPython.g4
```diff
@@ -1,39 +1,97 @@
-/**
- * Copyright 2021 The Aha001 Team.
+/*
```
wixette, line 1: For the LICENCE notice: I think we are not re-using this 3rd-party code. Actually, we are modifying it into a derived version, with a different license (given that the two licenses, i.e., Apache and MIT, are compatible). It's common to declare both the license of the derived work and the license of the original work in the file header. One of the best practices is:

```
Copyright 2021 The Aha001 Team

... The Apache License notice ...

This is a derived work of https://github.com/bkiers/Python3-parser.
Here is the original license notice:

Copyright (c) 2014 by Bart Kiers

... The original MIT license notice ...
```

See https://softwareengineering.stackexchange.com/questions/157968/how-to-manage-a-copyright-notice-in-an-open-source-project for an open discussion.

And, the Apache guidelines suggest a NOTICE file to list all the 3rd-party dependencies and their copyright and permission notices. See https://infra.apache.org/licensing-howto.html#permissive-deps . We can consider creating this NOTICE file later, since we have already included their source code notice.

grammars/SeedPython.g4
codingpotato: Done

wixette (approved):

PR #25, opened 2021-07-20
Update .editorconfig based on Google C# style

wixette:
See https://google.github.io/styleguide/csharp-style.html

- Names of classes, methods, enumerations, public fields, public properties, namespaces: PascalCase.
- Names of local variables, parameters: camelCase.
- Names of private, protected, internal and protected internal fields and properties: _camelCase.
- Naming convention is unaffected by modifiers such as const, static, readonly, etc.


codingpotato (approved):
LGTM

PR #27, opened 2021-07-20
A very basic code structure of SeedBlock

wixette:


csharp/src/SeedLang.Block/IEditable.cs
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace SeedLang.Block {
+  // All editable blocks should implement this interface to provide the content getter and setter to
+  // the UI layer.
+  public interface IEditable {
+    // Gets the editable block content, a plain text code.
```
codingpotato, line 19: Don't understand the meaning of "code". I guess it's the content of the editable block?

csharp/src/SeedLang.Block/NumberBlock.cs
```diff
@@ -0,0 +1,33 @@
...
+    public NumberBlock(string initialCode) {
+      UpdateCode(initialCode);
+    }
+
+    public override string GetEditableCode() {
+      return Value;
+    }
+
+    public override bool UpdateCode(string code) {
+      // TODO: Validates the input code by the underlying SeedLand engine.
```
codingpotato, line 27: Typo: SeedLand

codingpotato (changes requested):

csharp/src/SeedLang.Block/BlockProgram.cs
```diff
@@ -0,0 +1,62 @@
...
+          if (_instance == null) {
+            _instance = new BlockProgram();
+          }
+          return _instance;
+        }
+      }
+    }
+
+    BlockProgram() {
+      _modules = new List<Module>();
```
codingpotato, line 46: Is it possible to initialize it in the declaration statement?

codingpotato (changes requested):

csharp/src/SeedLang.Block/IEditable.cs
wixette: I intentionally implied the meaning of source code, since the blocks of SeedBlock are comparable to source code elements. I understand that this metaphor is a bit confusing here. Let's back off to the clearest term "text" - every editable block has an inner text to be edited.

csharp/src/SeedLang.Block/NumberBlock.cs
wixette: Done.

csharp/src/SeedLang.Block/BlockProgram.cs
wixette: Done

codingpotato (approved):
LGTM

PR #29, opened 2021-07-21
Implement three basic Ast nodes and visitor

codingpotato:
Implement number, string and binary expressions, eval statement node, Ast visitor and AstStringBuilder.


csharp/src/SeedLang.Ast/AstNode.cs
```diff
@@ -0,0 +1,24 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace SeedLang.Ast {
+  // The base class of all AST nodes.
+  public abstract class AstNode {
+    // Create the string representation of the AST node.
```
wixette, line 18: Although it's not mentioned by Google's or Microsoft's C# style doc, most Google coding styles enforce a descriptive style for file/class/function comments. For example:

> Google python style: The docstring should be descriptive-style ("""Fetches rows from a Bigtable.""") rather than imperative-style ("""Fetch rows from a Bigtable."""). 

> Google JavaScript style: Method descriptions begin with a verb phrase that describes what the method does. This phrase is not an imperative sentence, but instead is written in the third person, as if there is an implied This method ... before it.

I think either way is okay but we should better stick to one of them. How about following Google's convention to use the third-person phrases for file/class/function/method comments?

Here and the rest.

csharp/src/SeedLang.Ast/AstNode.cs
```diff
@@ -0,0 +1,24 @@
...
+    public override string ToString() => AstStringBuilder.AstToString(this);
```
wixette, line 19: Google C# style:

> Expression body syntax - Don’t use on method definitions. This will be reviewed when C# 7 is live, which uses this syntax heavily.

I think Google's decision is a bit outdated. We can make our own choice here. If we prefer the other way, please clarify this in the exception list in /CONTRUBUTING.md. Thanks.

Here and the rest.

csharp/src/SeedLang.Ast/AstNode.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    // Dispatch to the specific visit mothod of this node type.
```
wixette, line 21: s/mothod/method

csharp/src/SeedLang.Ast/AstVisitor.cs
```diff
@@ -0,0 +1,31 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace SeedLang.Ast {
+  // The base class of specialized visitor to traverse the AST tree.
+  public abstract class AstVisitor {
+    // Dispatch to the specific visit mothod of the node.
```
wixette, line 20: s/mothod/method

csharp/src/SeedLang.Ast/BinaryExpression.cs
```diff
@@ -0,0 +1,57 @@
...
+
+  public static class BinaryOperatorExtensions {
+    public static string Symbol(this BinaryOperator op) {
+      switch (op) {
+        case BinaryOperator.Add:
+          return "+";
+        case BinaryOperator.Substract:
+          return "-";
+        case BinaryOperator.Multiply:
+          return "*";
```
wixette, line 33: Since we plan to support multiple SeedX languages, how do we handle the case that two languages use different symbols for the same operator? E.g.,

Python uses `**` for the exponent operator. Lua uses `^` as the exponent operator.

Does the symbol in the AST module provide a common representation?

csharp/src/SeedLang.Ast/BinaryExpression.cs
```diff
@@ -0,0 +1,57 @@
...
+        case BinaryOperator.Divide:
```
wixette, line 34: Should we clarify that there are two division operators? E.g. Python's classical division / and floor division //.

csharp/src/SeedLang.Ast/AstNode.cs
codingpotato: Done

csharp/src/SeedLang.Ast/AstNode.cs
codingpotato: Done

csharp/src/SeedLang.Ast/AstVisitor.cs
codingpotato: Done

csharp/src/SeedLang.Ast/BinaryExpression.cs
codingpotato: Added Floor Divide, Power and Modulus operator

csharp/src/SeedLang.Ast/BinaryExpression.cs
codingpotato: This is a internal representation, only used in AST module. I changed it to internal and move it the file of AstStringBuilder.

wixette (approved):

PR #30, opened 2021-07-21
Initial implementation of Message and DiagnosticCollection

wixette:


csharp/src/SeedLang.Common/Diagnostic.cs
```diff
@@ -0,0 +1,38 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace SeedLang.Common {
+  // An immutable dignostic info, such as a compiler error or warning.
+  public class Diagnostic {
+    // The name or ID of the reporter.
+    public string Reporter { get; }
```
codingpotato, line 19: How about define a enum for all the reporters? Otherwise the developer may write wrong reporter string in the code easily.

csharp/src/SeedLang.Common/DiagnosticCollection.cs
```diff
@@ -0,0 +1,58 @@
...
+    public DiagnosticCollection() {
+    }
+
+    // Reports a new diagnostic.
+    public void Report(Diagnostic diagnostic) {
+      diagnostics.Add(diagnostic);
+    }
+
+    // Deletes a diagnostic from the collection.
+    public void Delete(Diagnostic diagnostic) {
```
codingpotato, line 44: Do we have requirement about deleting a diagnostic? Maybe the UI layer has strong requirement to search a diagnostic, like searching bya specific range.

csharp/src/SeedLang.Common/Severity.cs
```diff
@@ -0,0 +1,23 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace SeedLang.Common {
+  // Common severity levels for diagnostics, logging info, etc.
+  public enum Severity {
```
codingpotato, line 17: How about reversing the order, so that the big number indicates more serious error after serialization?

csharp/src/SeedLang.Common/Diagnostic.cs
wixette: The reporters of a diagnostic could be:

1. System components such as SeedAst Parser, Compiler, Interpreter, etc. These kinds of components can be pre-defined by an Enum type.
1. Third-party extensions. We need to allow third-party extensions to report diagnostics as well. For example, a visualizer extension subscribes to a visualizer provider. When the extension receives a set of data entries from the provider and figures out that there are logical inconsistent issues in the data set, it should have permission to report a diagnostic.

As a reference, VS Code uses a string URI for extensions to report diagnostics.

How about keeping the string type for reporters and providing a set of pre-defined string constants for the system components?

csharp/src/SeedLang.Common/DiagnosticCollection.cs
wixette: I agree. The deleting operations won't be necessary for most cases. I will remove those deleting operations.

As for the sorting/searching/indexing features, I will add clearer TODO comments and implement the features later.

csharp/src/SeedLang.Common/Severity.cs
wixette: Done. Also assigned explicit integer values to them.

codingpotato (approved):
LGTM

PR #33, opened 2021-07-25
Implement execution of binary expression and eval statement

codingpotato:


csharp/src/SeedLang.Ast/Executor.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace SeedLang.Ast {
+  // A executor class to execute an program represented by an AST tree.
+  //
+  // The executor traverses through the AST tree by implementing the interfaces of the AstVisitor
+  // class. The visit method of an expression will return the result of the expression. The return
+  // value of a statement visit method is not used.
+  public sealed class Executor : AstVisitor<BaseValue> {
+    // The global environment of the executor.
+    private readonly Dictionary<string, BaseValue> _globals = new Dictionary<string, BaseValue>();
```
wixette, line 26: Here the field "_globals" looks pretty like a global symbol table. It belongs to the Executor per this definition. This raises a good question:

Since there will be an AST executor as well as a byte code "executor" (or, the interpreter), will both the two executors share the common global symbol tables?

For example, the built-in (native or non-native) functions will be registered to the system. But do we need to register a built-in function twice, to the AST executor and to the byte code executor?

"print" will be a good case for this discussion. If the print function is implemented as a native function, do we need to register it twice?

I guess there can be a singleton LibraryManager (or named in other ways) class to hide all the complexities related to the library functions, including:

- Built-in functions implemented with SeedLang itself (e.g., implemented by SeedPython then compiled into an AST and a byte code)
- Built-in functions implemented with the native environment and expose their interfaces to SeedLan (e.g., implemented in C#)
- Safe third-party functions implemented with SeedLang itself.
- Unsafe third-party function implemented with the native environment, native DLLs, or even the native machine code.

Your thoughts?

csharp/src/SeedLang.Ast/Executor.cs
```diff
@@ -0,0 +1,78 @@
...
+    protected internal override BaseValue VisitNumberConstant(NumberConstantExpression number) {
+      return new NumberValue(number.Value);
+    }
+
+    protected internal override BaseValue VisitStringConstant(StringConstantExpression str) {
+      return new StringValue(str.Value);
+    }
+
+    protected internal override BaseValue VisitEvalStatement(EvalStatement eval) {
+      if (_globals["print"] is NativeFuncValue func) {
```
wixette, line 69: Is it the best choice to connect the built-in "print" function with the eval statement directly like this?

The semantic meaning of the eval statement is just "it results in a value" or "it returns a value", instead of "it prints a value to the standard output".

In a stack-based VM, a statement that returns a value simply leaves the value on the top of the stack. After that, the system may decide whether and how to use that value - a shell or a calculator may print the value out, while an interpreter may simply ignore the value and proceed to the next statement. Does a register-based VM share a similar behavior?

How about defining a consistent method to hold the return value of the eval statement somewhere in the VM? Hence, the calculator APP will have enough flexibility to make its own decision on how to run an executor and how to handle the result of an eval statement.

csharp/src/SeedLang.Ast/NumberValue.cs
```diff
@@ -0,0 +1,30 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace SeedLang.Ast {
+  // A immutable number value class.
```
wixette, line 18: A immutable -> An immutable

csharp/src/SeedLang.Ast/AstStringBuilder.cs
```diff
@@ -56,20 +63,24 @@ internal sealed class AstStringBuilder : AstVisitor {
       return asb.ToString();
     }
 
-    protected internal override void VisitBinaryExpression(BinaryExpression binary) {
+    protected internal override NullResult VisitBinaryExpression(BinaryExpression binary) {
       _out.Append($"({binary.Left} {binary.Op.Symbol()} {binary.Right})");
+      return null;
```
wixette, line 68: The combination of `NullResult VisitFunc(...)` and `return null` is really confusing to me. Here I think "NullResult" intends to imply "no result will be returned", instead of "it may return a null result to indicate something interesting".

It's okay to use generic patterns to cover a function with different return types. But It sounds a bit weird to me to use a generic pattern to cover both the function's return types and the case that the function returns nothing.

How about the following way:

```
class NoResult {
  public const NoResult Nothing=>null;
}

// For the visitor functions that return a value:
NumberValue VisitBinaryExpression(...)

// For the visitor functions that do not return any value:
NoResult VisitBinaryExpression(...; return NoResult.Nothing; )
```

Or, I think we can also consider an OO solution, like:

```
class Result {...}
class NoResult : Result {...}
class BaseValue : Result {...}
class NumberValue : BaseValue {...}

// Then the base class of all visitors can define the abstract methods like:
Result VisitBinaryExpression(...);

// The concrete visitor with no result could define the method as:
NoResult VisitBinaryExpression(...; return NoResult.Nothing;
```

codingpotato: The difficulty is void or Void is not allowed to be used as template parameter in C#. But it's allowed in C++. There are some consideration in this design:

1. The results of visit methods are needed, otherwise the visitor need maintain complex states during the AST tree tranversal.
2. The AstVisitor can be used by several clients. For example, the compiler in the interpreter will use it to travers the AST tree to convert it to bytecodes. In this situation, the return value might be a class defined bby the compiler. I don't think there is any relationship between this class and the BaseValue. So I think generic result type can express it well.

So maybe use your first solution now. We can change it later if we have better idea.

csharp/src/SeedLang.Ast/Executor.cs
```diff
@@ -0,0 +1,78 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+
+namespace SeedLang.Ast {
+  // A executor class to execute an program represented by an AST tree.
```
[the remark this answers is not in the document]
codingpotato: Done

csharp/src/SeedLang.Ast/Executor.cs
codingpotato: 1. The AST executor and VM executor can not share the common global symbol table, because they use the different value types because of the performance consideration. But the build-in functions they use might be the same ones.
2. The design of following senaria might be:
- Build-in functions implemented with SeedLang itself: they shall be AST trees in the AST executor, and will be compiled to bytecode and stored in a function value of VM.
- Build-in functions implemented with the native enviorment: I considered to create a sandbox class, and store them in the sandbox object. Both AST executor and VM executor will call into the sandbox object.
- Third-party functions implemented with SeedLang itself: shall be the same as the SeedLang build-in functions.
- Unsafe third-party functions implemented with the native enviorment: shall be the same as build-in native functions.

In current phase, I decided to remove the global symbol table, and expose the execution result through a visualizer.

csharp/src/SeedLang.Ast/Executor.cs
codingpotato: Remove the print native function and expose the result through a visualizer.

csharp/src/SeedLang.Ast/NumberValue.cs
codingpotato: Done

csharp/src/SeedLang.Ast/NativeFuncValue.cs
```diff
@@ -0,0 +1,35 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace SeedLang.Ast {
+  // A immutable native function class.
```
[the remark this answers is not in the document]
codingpotato: This file os removed.

csharp/src/SeedLang.Ast/Executor.cs
wixette: Agreed, and just one more clarification: the concept "sandbox" covers two things:

1. As your described, the "sandbox" component runs the built-in and third-party unsafe native extensions and makes sure that the extensions do not violate the safe boundary.
2. Event if a third-party extension is implemented in SeedLang itself, it still requires a sandbox layer (probably a much thinner layer) to protect - for example, attackers may implement an infinite loop to hang the hosting service. The thin sandbox layer should be able to detect such behaviors and solve them.

csharp/src/SeedLang.Ast/AstStringBuilder.cs
wixette: Agreed.

wixette (approved):

PR #34, opened 2021-07-31
Implement CodeRange and complete DiagnosticCollection

wixette:


wixette:
@codingpotato I simplified the class structure of the position classes and the range classes. Please take another look. Thanks,

csharp/src/SeedLang.Common/BlockPosition.cs
```diff
@@ -0,0 +1,55 @@
...
+    public override string ToString() {
+      return $"Block: {BlockId}";
+    }
+
+    public bool Equals(BlockPosition pos) {
+      return !(pos is null) && BlockId == pos.BlockId;
+    }
+
+    public override bool Equals(object obj) {
+      return (obj is BlockPosition objBlockPosition) && Equals(objBlockPosition);
```
codingpotato, line 41: The code can be "this.Equals(obj as BlockPosition);" I tried another implement "return false;", which passes all the test cases as well. I think the reason is if obj is the type of BlockPosition, it will call the type specific version. This version will handle all cases of non BlockPosition type. It's true if the BlockPosition is sealed without subclass.

csharp/src/SeedLang.Common/BlockRange.cs
```diff
@@ -0,0 +1,49 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace SeedLang.Common {
+  // Represents a range in a SeedBlock source code.
+  public class BlockRange : Range {
+    public static readonly BlockRange Empty = new BlockRange(null as BlockPosition);
```
codingpotato, line 18: I understand the code here is using null as a null BlockPosition. If we didn't need distinguish Null BlockRange and null TextRange, we can have a NullRange class implement Range to indicate it cannot be associated with any code position. Then The BlockPosition and TextPosition can be struct without null. It can remove the null check for position objects.

csharp/src/SeedLang.Common/Range.cs
```diff
@@ -12,10 +12,34 @@
...
-    // Todo: implement the Range class.
+  // The base class of all the concrete code range classes.
+  public abstract class Range : IEquatable<Range> {
+    // Returns if the range is empty. A range can be empty when a diagnostic cannot be associated to
+    // a particular code position.
+    public abstract bool IsEmpty();
+
+    public abstract override int GetHashCode();
+
+    public abstract override string ToString();
```
codingpotato, line 26: The GetHashCode and ToString are already defined in Object. I guess The intention to redefine it here is to hide the default implement and make sure all subclasses implement it. Then it's fine.

csharp/src/SeedLang.Common/TextPosition.cs
```diff
@@ -0,0 +1,102 @@
...
+    public override int GetHashCode() {
+      return Tuple.Create(Line, Column).GetHashCode();
+    }
+
+    public override string ToString() {
+      return $"Ln {Line}, Col {Column}";
+    }
+
+    public int CompareTo(TextPosition pos) {
+      if (pos is null) {
```
codingpotato, line 37: If it's a struct, all the null check in this classes can be removed safely.

csharp/src/SeedLang.Common/BlockPosition.cs
```diff
@@ -0,0 +1,55 @@
...
+    public override int GetHashCode() {
+      return BlockId.GetHashCode();
+    }
+
+    public override string ToString() {
+      return $"Block: {BlockId}";
+    }
+
+    public bool Equals(BlockPosition pos) {
+      return !(pos is null) && BlockId == pos.BlockId;
```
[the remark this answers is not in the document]
wixette: I've made BlockPosition and TextPosition structs as you suggested, following the struct example of the "How to define value equality for a type" doc.

csharp/src/SeedLang.Common/BlockPosition.cs
wixette: Let's follow the "How to define value equality for a type" doc here.

csharp/src/SeedLang.Common/BlockRange.cs
wixette: I double-checked on this and realized there is not urgent request on a NullRange or an EmptyRange for now. We can simply pass a null reference to the new Diagnostic instance as the Range argument. No need to maintain a special instance or a special subclass.

csharp/src/SeedLang.Common/Range.cs
wixette: GetHashCode() is for IEquality<>. The reason that I implement IEquality in the base class (the equality implementation of the Range/TextRange/BlockRange classes are slightly different from the example given by [How to define value equality for a type](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/how-to-define-value-equality-for-a-type#class-example). The subclass like TextRange and BlockRange still accept a base class type as its argument of Equals. This provides the flexibility that we can compare the instances of the two different sub-types with "==" or "!=".

csharp/src/SeedLang.Common/TextPosition.cs
wixette: I've made BlockPosition and TextPosition structs as you suggested

csharp/src/SeedLang.Common/BlockPosition.cs
wixette: In the updated code: class equality (like in class Range) uses "Equals(obj as Range)". struct equality (like in TextPosition) still uses "(obj is TextPosition ...)" since struct types are not nullable.

csharp/src/SeedLang.Common/Range.cs
wixette: ToString() is kept to hide the default implement as enforce all subclasses to provide a customized implementation.

codingpotato (approved):

PR #35, opened 2021-08-01
Implement SeedLang.Shell to parse statements

codingpotato:
A very simple interactive SeedLang shell.

- Run SeedLang.Shell
- Input a statement
- If it's a valid statement, shows the parsed AST, otherwise shows parse errors.


wixette (commented):
The DignosticCollection part in this pr is conflicting with my pr https://github.com/aha-001/SeedLang/pull/34. Please take a look and let me know your thoughts. Thanks,

codingpotato:
Merged main branch and used the new DiagnosticCollection interface. Please have a review again. Thanks.

csharp/src/SeedLang.X/PythonParser.cs
```diff
@@ -23,24 +24,25 @@ namespace SeedLang.X {
...
-      } catch (Exception) {
-        // TODO: implement parse exception handling.
-        throw;
+    public static AstNode Parse(string source, ParseRule rule, DiagnosticCollection diagnostics) {
+      var inputStream = new AntlrInputStream(source);
+      var lexer = new SeedPythonLexer(inputStream);
+      var tokenStream = new CommonTokenStream(lexer);
+      var parser = new SeedPythonParser(tokenStream);
+
+      // Remove default error listerners which include a console error reporter.
```
wixette, line 33: Remove -> Removes

csharp/src/SeedLang/Engine.cs
```diff
@@ -0,0 +1,48 @@
...
+      }
+    }
+
+    private Engine() {
+    }
+
+    // Runs a statement and returns diagnostic information collected during parsing and execution.
+    public static DiagnosticCollection RunStatement(string source) {
+      var collection = new DiagnosticCollection();
+      AstNode node = PythonParser.Parse(source, ParseRule.Statement, collection);
```
wixette, line 40: Since the Engine is generic for all SeedLang languages, is it good to hard-code a PythonParser here? Adding TODO comments here should be okay, and we can make this line a general logic later.

csharp/src/SeedLang.X/PythonParser.cs
```diff
@@ -23,24 +24,25 @@ namespace SeedLang.X {
...
+      parser.AddErrorListener(errorListener);
+      var visitor = new PythonVisitor();
+
+      switch (rule) {
+        case ParseRule.Expression:
+          return visitor.Visit(parser.expr());
+        case ParseRule.Statement:
+          return visitor.Visit(parser.stmt());
+        default:
+          throw new ArgumentException("Unknown parse rule: " + rule);
```
wixette, line 45: Do we need to catch this exception (and, others that are thrown by underlying components) in the shell's main loop?

And, if the message string of this exception will be seen by end-users, it should be localized by the Common.Message class. If so, please switch to the Common.Message solution or at least add TODO comments here.

csharp/src/SeedLang/Engine.cs
codingpotato: Add TODO.

csharp/src/SeedLang.X/PythonParser.cs
codingpotato: Done.

csharp/src/SeedLang.X/PythonParser.cs
codingpotato: This default case wull never happen, because the value is out of the range of the enum values. We can have several options:
1. Use Debug.Assert() and let the program terminate if it really happen (not possible)
2. Use InvalidEnumArgumentException, and let the program terminate.
3. Use InvalidEnumArgumentException, and catch the exception in the shell's main loop. But what we can do is printing the exception. It's as same as the default behavior. 
What's your thought about it?

wixette: If the case would not happen, either one of your choices is okay I think.

wixette (approved):

PR #40, opened 2021-08-08
Define the IEngine interface

codingpotato:
Define the IEngine interface. Refine the SeedPython grammar file and the SeedPython parser to validate and parse a single expression and statement.

### DIFF
--- a/csharp/src/SeedLang.Common/Language.cs
+++ b/csharp/src/SeedLang.Common/Language.cs
@@ -13,12 +13,9 @@
 // limitations under the License.
 
 namespace SeedLang.Common {
-  // The parse rules which can be used to validate and parse SeedBlock or SeedX languages.
-  public enum ParseRule {
-    Identifier,
-    Number,
-    String,
-    Expression,
-    Statement,
+  // All languages that the SeedLang system supports.
+  public enum Language {
+    Block,
+    Python,
   }
 }
--- a/csharp/src/SeedLang.Common/RunType.cs
+++ b/csharp/src/SeedLang.Common/RunType.cs
@@ -13,12 +13,13 @@
 // limitations under the License.
 
 namespace SeedLang.Common {
-  // The parse rules which can be used to validate and parse SeedBlock or SeedX languages.
-  public enum ParseRule {
-    Identifier,
-    Number,
-    String,
-    Expression,
-    Statement,
+  // The running type of SeedBlock and SeedX source code.
+  public enum RunType {
+    // Parses and validates the source code without runing.
+    Dryrun,
+    // Parses the source code into an AST tree, and runs it by traversing the AST tree.
+    Source,
+    // Parses and compiles the source code into bytecode, and runs it in a VM.
+    Bytecode,
   }
 }
--- a/csharp/src/SeedLang.Common/RunType.cs
+++ b/csharp/src/SeedLang.Common/RunType.cs
@@ -16,9 +16,9 @@ namespace SeedLang.Common {
   // The running type of SeedBlock and SeedX source code.
   public enum RunType {
     // Parses and validates the source code without runing.
-    Dryrun,
+    DryRun,
     // Parses the source code into an AST tree, and runs it by traversing the AST tree.
-    Source,
+    Ast,
     // Parses and compiles the source code into bytecode, and runs it in a VM.
     Bytecode,
   }
--- a/csharp/src/SeedLang.Shell/Program.cs
+++ b/csharp/src/SeedLang.Shell/Program.cs
@@ -13,11 +13,40 @@
 // limitations under the License.
 
 using System;
+using CommandLine;
+using CommandLine.Text;
 using SeedLang.Common;
 
 namespace SeedLang.Shell {
   class Program {
-    static void Main(string[] _) {
+    class Options {
+      [Option('r', "rule", Required = false, Default = "statement",
+        HelpText = "The parse rule of the source code."
+                   + " Allowed values are statement, expression, number or identifier.")]
+      public string Rule { get; set; }
+
+      [Option('l', "language", Required = false, Default = "python",
+        HelpText =
+          "The language of the source code. Allowed values are python, lua or javascript.")]
+      public string Language { get; set; }
+
+      [Option('d', "dryrun", Required = false,
+        HelpText = "Verify the source code without running it.")]
+      public bool Dryrun { get; set; }
+    }
+
+    static void Main(string[] args) {
+      var parserResult = Parser.Default.ParseArguments<Options>(args);
+      var helpText = HelpText.AutoBuild<Options>(parserResult, h => h, e => e);
+      parserResult.WithParsed(options => {
+        Console.Error.WriteLine(helpText.Heading);
+        Console.Error.WriteLine(helpText.Copyright);
+        Console.Error.WriteLine();
+        RunOptions(options);
+      });
+    }
+
+    private static void RunOptions(Options options) {
       // TODO: add a REPL class to encapsulate the input and execution process.
       while (true) {
         Console.Write("> ");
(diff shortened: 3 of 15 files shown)


csharp/src/SeedLang.Common/RunType.cs
```diff
@@ -0,0 +1,25 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace SeedLang.Common {
+  // The running type of SeedBlock and SeedX source code.
+  public enum RunType {
+    // Parses and validates the source code without runing.
+    Dryrun,
+    // Parses the source code into an AST tree, and runs it by traversing the AST tree.
+    Source,
```
wixette, line 21: Which one is better, Source or AST? I slightly incline to AST. Just ask this question for you to double-check.

csharp/src/SeedLang.Shell/Program.cs
```diff
@@ -13,20 +13,83 @@
...
+using CommandLine.Text;
 using SeedLang.Common;
 
 namespace SeedLang.Shell {
   class Program {
-    static void Main(string[] _) {
+    class Options {
+      [Option('r', "rule", Required = false, Default = "statement",
+        HelpText = "The parse rule of the source code." +
+                   " Allowed values are identifier, number, string, expression or statement.")]
```
wixette, line 25: Optional: consider to auto-generate this allowed value list from enum.GetName().

csharp/src/SeedLang.Shell/Program.cs
```diff
@@ -13,20 +13,83 @@
...
+      public string Rule { get; set; }
+
+      [Option('l', "language", Required = false, Default = "python",
+        HelpText =
+          "The language of the source code. Allowed values are python, lua or javascript.")]
+      public string Language { get; set; }
+
+      [Option('t', "type", Required = false, Default = "source",
+        HelpText = "Run type. Allowed values are dryrun, source or bytecode.")]
```
wixette, line 34: Optional: consider to auto-generate this type list from enum.GetName().

csharp/src/SeedLang.Shell/Program.cs
```diff
@@ -13,20 +13,83 @@
...
+        RunOptions(options);
+      });
+    }
+
+    private static void RunOptions(Options options) {
       // TODO: add a REPL class to encapsulate the input and execution process.
+      Language language = Language.Python;
+      ParseRule rule = ParseRule.Statement;
+      switch (options.Rule) {
+        case "identifier":
```
wixette, line 54: Consider case enum.GetName(ParseRule.Identifier).ToLower() ?

Here  and the rest.

csharp/src/SeedLang/IEngine.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace SeedLang {
+  // The interface of the SeedLang engine.
+  public interface IEngine {
+    // A collection that contains all binary operators.
+    IEnumerable<string> BinaryOperators { get; }
+
+    // Runs source code with the specified module name, language, target parsing rule and runing
+    // type.
+    //
+    // There are three mothods to run the source code.
```
wixette, line 27: mothods -> methods

csharp/src/SeedLang/IEngine.cs
```diff
@@ -0,0 +1,37 @@
...
+    // 1) Dryrun: parses and validates the source code without runing.
+    // 2) Source: parses the source code into an AST tree, and runs it by traversing the AST tree.
+    // 3) Bytecode: parses and compiles the source code into bytecode, and runs it in a VM.
+    //
+    // The parsing and running diagnostic information will be collected into the diagnostic
+    // collection.
+    bool Run(string source, string module, Language language, ParseRule rule, RunType runType,
```
wixette, line 34: We have decided via emails that the dependency chain will be:

Shell / Unity UI -> Engine -> parsers implemented by SeedBlock / SeedX -> SeedAst

Meanwhile, there will obviously be another dependency chain for the UI to operate on in-memory blocks:

Unity UI -> SeedBlock

Given the above dependency chains, a Unity UI can parse a SeedBlock code into an Engine instance to parse the code. Hence, the interface Run() here need to accept an in-memory SeedBlock module object and parse it to SeedBlock's parser.

It results in a new cycle dependency if we keep the things as they are.

Probably we need to extract the interface of the in-memory block (the BaseBlock interface) out to SeedLang.Common?

Your thoughts?

csharp/src/SeedLang/IEngine.cs
codingpotato: Done

csharp/src/SeedLang/IEngine.cs
codingpotato: The dependency should be as following:

```
                                 Unity UI
                          -----------------
                          |                            |
                          |                            |
                         V                          V
               SeedBlock. <------ Engine -------> X
                         |                                                     |
                         |                                                     |
                        --------------------------------
                                                |
                                               V
                                             Ast

```

csharp/src/SeedLang.Common/Language.cs
```diff
@@ -0,0 +1,21 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace SeedLang.Common {
+  // All languages that the SeedLang system supports.
+  public enum Language {
```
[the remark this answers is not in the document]
codingpotato: Done

csharp/src/SeedLang.Common/RunType.cs
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace SeedLang.Common {
+  // The running type of SeedBlock and SeedX source code.
+  public enum RunType {
+    // Parses and validates the source code without runing.
+    Dryrun,
```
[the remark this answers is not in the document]
codingpotato: Done

csharp/src/SeedLang.Common/RunType.cs
codingpotato: I used "source", becase I'm not sure if the AST runing is meaningful for the user. Maybe we can keep AST now, and see if it needs be changed later.

csharp/src/SeedLang.Shell/Program.cs
codingpotato: The attribute argument must be constant expression. Keep it now, and see if I can find a better way later.

csharp/src/SeedLang/IEngine.cs
wixette: Ah... my bad... I mistakenly mixed something in my mind when I was reading your code. Your diagram is pretty clear. Forget my comments.

csharp/src/SeedLang.Shell/Program.cs
codingpotato: The case of the switch statement must be constant. I added a TODO here to consider using dictionary later.

csharp/src/SeedLang/IEngine.cs
```diff
@@ -0,0 +1,37 @@
...
+// limitations under the License.
+
+using System.Collections.Generic;
+using SeedLang.Common;
+
+namespace SeedLang {
+  // The interface of the SeedLang engine.
+  public interface IEngine {
+    // A collection that contains all binary operators.
+    IEnumerable<string> BinaryOperators { get; }
```
wixette, line 22: Since we have re-defined the dependency graph, I think there is no need for client code (Unity UI) to get the list of supported binary operators from IEngine directly:

Yes, Unity UI will need to know all the supported operators but it can also talk to SeedLang.Block to get the list. And, since SeedLang.Block will have its PartialParser and FullParser inside the assembly boundary, its PartialParser and FullParser must have the knowledge of supported binary operators (e.g., they can get that knowledge from the ANTLR generated code based on grammars/seed_block.g4). Hence, Unity UI can surely get the operator list from SeedLang.Block, right?

Given this, I think it's better not to expose this interface from IEngine.

My original request of this interface was based on my previous idea that SeedLang.Block depends on IEngine. The request is obsolete now since the dependency does not exist any more.

codingpotato: Sure. I will remove this interface now. And I'm going to add the Block parser in the next two or three PRs. We can see if it's easy to get the operators from SeedBlock parser.

wixette (approved):

PR #41, opened 2021-08-09
Initial design doc of SeedBlock

wixette:
* The initial design doc of SeedBlock - the doc is incomplete but it already covers some critical contents (such as the positioning system and the block categories) that we require for building a basic SeedLang.Block module.
* Update the design dir to use SVG images (instead of PNG images) for better rendering quality.
* Add a README to the design dir.
* Add the yarn support to the block proto util (the util is used to generate SVG example images for the design doc).

See https://github.com/aha-001/SeedLang/blob/wip_wixette_seedblockdesign/design/seed_block.md for the rendered design doc.


codingpotato:
The SeedBlock doc is good. Just double check my understanding here:

For example in the "set 'count' to 'x + 3.14'" block, there are two input slots in this block. Only one expression block can be docked to one input slot. Is it right?
If I want to connect a expression "x + y * z" to the second input slot, I just need one expression block with 5 items "x", "+", "y", "*", "z" in it, no need two nested binary expression blocks? If so, the user need dock these 5 items to the expression block. Do we need define the dock point of this expression block for these 7 items? Or these 5 items are not defined as blocks? They just can be edit by the inline editor? May be we need clarify it in this doc.

In this example, the "count" is the left value, and the "x + 3.14" is the right value. Is there ant restriction which expression blocks can be docked as left values, and which expression blocks can be docked as right values?

codingpotato (commented):
Please see my comment in above conversation.

wixette:
> The SeedBlock doc is good. Just double check my understanding here:
> 
> For example in the "set 'count' to 'x + 3.14'" block, there are two input slots in this block. Only one expression block can be docked to one input slot. Is it right?
> If I want to connect a expression "x + y * z" to the second input slot, I just need one expression block with 5 items "x", "+", "y", "*", "z" in it, no need two nested binary expression blocks? If so, the user need dock these 5 items to the expression block. Do we need define the dock point of this expression block for these 5 items? Or these 5 items are not defined as blocks? They just can be edit by the inline editor? May be we need clarify it in this doc.
> 
> In this example, the "count" is the left value, and the "x + 3.14" is the right value. Is there ant restriction which expression blocks can be docked as left values, and which expression blocks can be docked as right values?

According to my idea, the "expression block" is an open container that accepts an unlimited open list of Input Slots. So for your example, "x + y * z", users can simply drag blocks like "x", "+", "y" one by one to the expression block, as a flat list of Input blocks. No need to organize them in AST like structure. Users can insert a block (by drag-and-drop) into a middle position (although, I haven't implement this logic yet) of the list, or delete some block from the list. Of course, users can open the inline editing box to input a raw string to form the expression. In the inline editing mode, the partial parser inside SeedBlock will be responsible to parse the raw string and generate blocks like "x", "+", "y".

I think this design gives users the largest space of freedom to build or organize an expression block on the block canvas or via the inline editing box.

As for the left value and the right value, I think we can make a clear definition at the language's semantical level. Blocks or block groups like "counter (variable)", "someModule.property (dot access expression)" may result in left value, and can be dragged and dropped to the first Input slot of the "set ... to ..." block.

As a consequence, the partial parser in the SeedBlock will need to answer this question: given an expression block type, is the block resulting in a left value or a right value? 

Let me know if you have better ideas. And, this PR is just an initial and incomplete edition of the SeedBlock design. I will clarify more on the above thoughts to the design doc, via future PRs.

codingpotato (approved):
@wixette Agree with you. We may need define a name for the drop point of the expression block and state in this doc later.

The doc looks good to me now.

PR #43, opened 2021-08-14
Implement syntax errors collection

codingpotato:


csharp/src/SeedLang.Common/Message.cs
```diff
@@ -68,6 +68,11 @@ public enum Message {
     InvalidJson1,                       // Not a valid JSON string/file.
     InvalidOperatorName1,               // Invalid operator name.
     InvalidPrimitiveValue1,             // Invalid primitive value.
+    SyntaxErrorFailedPredicate1,        // Semantic predicate failed syntax error.
+    SyntaxErrorInputMismatch2,          // Input mismatch syntax error.
+    SyntaxErrorMissingToken2,           // Missing token syntax error.
+    SyntaxErrorNoViableAlternative1,    // No viable alternative path syntax error.
+    SyntaxErrorUnwantedToken2,          // Unwanted token syntax error.
```
wixette, line 75: I like this "SyntaxError" prefix since it is a general error type name, instead of a component name. These kinds of prefixes would somehow solve the case of sharing common messages among components.

Let's try if all the other messages can be prefixed with this convention.

wixette (approved):

PR #44, opened 2021-08-16
Evaluate SeedPython statements and visualize the result

codingpotato:


csharp/src/SeedLang.Runtime/EventHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+  internal interface ISubscribable<in Visualizer> {
+    void Subscribe(Visualizer visualizer);
+    void Unsubscribe(Visualizer visualizer);
+  }
+
+  // The event handler to notify all the subscribed visualizers when the event is triggered.
+  public class EventHandler<Event> : ISubscribable<IVisualizer<Event>> {
+    private readonly List<IVisualizer<Event>> _visualizers = new List<IVisualizer<Event>>();
+
+    public void Subscribe(IVisualizer<Event> visualizer) {
```
wixette, line 28: The names look confusing to me.

> **Subscribe**: arrange to receive something, typically a publication, regularly by paying in advance.
"subscribe to the magazine for twelve months and receive a free limited-edition T-shirt"

In our context, it's okay to say "an event receiver (or event handler) subscribes to an event type published by an event publisher", where:

* Event receiver/handler -> a concrete visualizer
* Event type -> event type
* Event publisher -> the class that sends the event to the visualizer

A couple of confusing things here:

* You named this class "EventHander". This class is actually sending events to the registered visualizers. So this class is just like an event publisher, while the concrete visualizers are more like "event handlers" than this class.
* "Subscribe" is mostly used in the context that "a consumer subscribes to a message type". Here, we actually want "a visualizer subscribes to an event type". But the current code reads "an event publisher subscribes to a visualizer(event handler)".

See [Kafka's subscribe API](https://kafka.apache.org/28/javadoc/org/apache/kafka/clients/consumer/Consumer.html#subscribe(java.util.Collection) for a reference to define a subscribe method from a consumer side:

```
Consumer.subscribe
void subscribe​(Collection<String> topics)
```

From a publisher side, similar functionalities can be defined as:

```
Publisher.addEventListener(...);
Publisher.removeEventListener(...);
```

csharp/src/SeedLang.Runtime/EventHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+      _visualizers.Add(visualizer);
```
wixette, line 29: Do we need to check if the same visualizer is registered more than once?

csharp/src/SeedLang.Runtime/VisualizerCenter.cs
```diff
@@ -0,0 +1,57 @@
...
+        if (property.PropertyType.IsGenericType &&
+            property.PropertyType.GetGenericTypeDefinition() == typeof(EventHandler<>)) {
+          _eventHandlers.Add(property.GetValue(this));
+        }
+      }
+    }
+
+    // Subscribes a visualizer to this visualizer center. Loops each event handler, and subscribe to
+    // it if the visualizer implement the IVisualizer interface of this event.
+    public void Subscribe<Visualizer>(Visualizer visualizer) {
```
wixette, line 39: Again, it's really confusing to say "a visualizer center subscribes to a visualizer".

Instead, we'd better say "a visualizer subscribes to an event type publishing by a center or a dispatcher".

Or, we can also say "a center or a dispatcher or an event source adds a visualizer as its listener for a certain event type".

wixette (changes requested):

csharp/src/SeedLang.Ast/Executor.cs
```diff
@@ -18,17 +18,14 @@
 namespace SeedLang.Ast {
   // An executor class to execute a program represented by an AST tree.
   public sealed class Executor : AstWalker {
-    // The visualizer object to visualize the execution result.
-    private readonly IVisualizer _visualizer;
+    // The visualizer center to observer AST execution events and dispatch to the subscribed
```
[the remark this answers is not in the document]
codingpotato: Done

csharp/src/SeedLang.Ast/Expressions.cs
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System;
+using SeedLang.Runtime;
```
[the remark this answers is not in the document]
codingpotato: 1. There will be only one SeedLang.Runtime component. All the common runtime classes include Value interface, Visualizer Center, and maybe Sandbox etc. will be defined in this component. Other runtime classes used specially in AST execution and VM execution will be located in SeedLang.Ast or SeedLang.VM (I'm thinking about renaming it to SeedLang.Interpreter. VM is just a class in this component).
2. I defined RunType.DryRun, RunType.Ast and RunType.Bytecode in the common component. This enum is only used in the Engine and not needed in the SeedLang.Runtime component.
3. The SeedLang.Ast and SeedLang.VM will define their own Value classes and implement the IValue interface from SeedLang.Runtime component. They also use Visualizer and Sandbox class from Runtime component. So the dependency graph is Ast -> Runtime, VM -> Runtime, instead of the opposite direction.

csharp/src/SeedLang.Runtime/VisualizerCenter.cs
```diff
@@ -0,0 +1,57 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Collections.Generic;
+
+namespace SeedLang.Runtime {
+  // The visualizer center to observer execution events and dispatch them to the subscribed
```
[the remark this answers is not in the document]
codingpotato: Done

csharp/src/SeedLang.Runtime/EventHandler.cs
codingpotato: I refer to the naming of Microsoft observer design pattern: [https://docs.microsoft.com/en-us/dotnet/standard/events/observer-design-pattern](https://docs.microsoft.com/en-us/dotnet/standard/events/observer-design-pattern). Maybe different person has dofference interpreter of this naming.
But I think your suggestion about publisher might be better. I will try to change the naming accordingly.

csharp/src/SeedLang.Runtime/EventHandler.cs
codingpotato: Add the same check.

csharp/src/SeedLang.Ast/Expressions.cs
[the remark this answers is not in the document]
wixette: This makes sense to me. Thanks,

csharp/src/SeedLang.Runtime/EventHandler.cs
wixette: I guess Microsoft doesn't have a good reputation in the API design area. Considering the following code:

The example code at https://docs.microsoft.com/en-us/dotnet/api/system.iobservable-1?view=net-5.0

```
public class LocationReporter : IObserver<Location>
{
  public virtual void Subscribe(IObservable<Location> provider)
   {
      if (provider != null)
         unsubscriber = provider.Subscribe(this);
   }
}
```

See? Microsoft uses the name "Subscribe" twice, with totally different meanings :-) In Microsoft's logic, an observer can "subscribe" to an observable object, and, on the other hand, an observable object (provider) can also "subscribe" to an observer. Super confusing!

In _Design Patterns_ there is a subscriber example under the pattern **Observer**. See "Example - News Agency at "https://www.oodesign.com/observer-pattern.html. The registration action is initiated by the observable object (or the publisher), but the methods name are "attach" and "detach". This name pair is way better than Microsoft's solution.

So, I think there are a couple of good practices we can follow:

A) Message queue like pub-sub pattern:

```
Publisher.register(eventTypeOrTopic)
Publisher.unregister(eventTypeOrTopic)
Consumer.subscribe(publisher, eventTypeOrTopic)
Consumer.unsubscribe(publisher, eventTypeOrTopic)
```

B) JavaScript like event-listener pattern:

```
Element.dispatchEvent(eventType);
Element.addEventListener(eventHandler);
Element.removeEventListener(eventHandler);
```

C) _Design Patterns_'s observer pattern:

```
Publisher.attach(subscriber);
Publisher.detach(subscriber);
Publisher.notifyObservers();
Subscriber.update();
```

If our Visualization mechanism will be used in both sync and async circumstances, I'd equally prefer choice A) or choice B). Otherwise, if we only consider the sync case, I'd prefer choice C).

codingpotato: Choose A)

csharp/src/SeedLang.Runtime/IVisualizer.cs
```diff
@@ -13,12 +13,8 @@
...
 namespace SeedLang.Runtime {
-  // The interface of all visualizer classes.
-  public interface IVisualizer {
-    // Called when a binary expression is evaluated.
-    void OnBinaryExpression(IValue left, IValue right, IValue result);
-
-    // Called when an eval statement is executed.
-    void OnEvalStatement(IValue value);
+  // The interface of visualizers to visualize the given event.
+  public interface IVisualizer<in Event> {
```
wixette, line 17: I tried to remove the contravariance modifier "in" from this line - all the unit tests still passed. Are there any special cases that depend on this contravariance generic parameter? If so, it'd be good to add a unit test to cover that.

The other case is on the contrary side: the contravariance modifier "in" used by `IPublisher<in Visualizer>` is really required, otherwise the unit tests will fail. (It is required by `p is IPublisher<Visualizer> publisher` I think).

codingpotato: I will remove the modifier "in" here, because I don't plan to make event classes polymorphic. So the Event parameter doesn't need to be contravariance.
Yes, the contravariance modifier "in" used by `IPublisher<in Visualizer>` is really required. I will add a comment before `p is Publisher<Visualizer> publisher` check to make it clear in the future.

wixette (approved):

Issue #45, opened 2021-08-19
null value from ANTLR is not handled, for expressions like "eval x" or "eval x*2"


wixette:
1) Start SeedLang.Shell, enter "eval x\n":

The returned node of parser.single_stmt() contains a null field: node.Expr = null. Meanwhile localCollection.Diagnostics.Count is still 0. The outer logic treated this as a normal case and continued the execution. The output from the shell is:

```
eval
```

2) Start SeedLang.Shell, enter "eval x+2\n":

Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at SeedLang.Ast.BaseValue.op_Addition(BaseValue lhs, BaseValue rhs) in /Users/ygwang/src/SeedLang/csharp/src/SeedLang.Ast/BaseValue.cs:line 25

codingpotato:
Fixed by commit #47

PR #47, opened 2021-08-21
Parse and execute the identifier expression and assignment statement

codingpotato:
Fix issue #45

### DIFF
--- a/csharp/src/SeedLang.Ast/Executor.cs
+++ b/csharp/src/SeedLang.Ast/Executor.cs
@@ -24,6 +24,7 @@ namespace SeedLang.Ast {
     // visualizers.
     private readonly VisualizerCenter _visualizerCenter;
 
+    // The dictionary to store variable names and current values of global variables.
     private readonly Dictionary<string, BaseValue> _globals = new Dictionary<string, BaseValue>();
 
     // The result of current executed expression.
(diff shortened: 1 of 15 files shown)


csharp/src/SeedLang.Ast/Executor.cs
```diff
@@ -21,6 +23,10 @@ public sealed class Executor : AstWalker {
     // The visualizer center to observe AST execution events and dispatch them to the registered
     // visualizers.
     private readonly VisualizerCenter _visualizerCenter;
+
+    // The dictionary to store variable names and current values of global variables.
+    private readonly Dictionary<string, BaseValue> _globals = new Dictionary<string, BaseValue>();
```
wixette, line 28: Should we have a common class hierarchy for symbol tables, located in SeedLang.Runtime? Will the symbol table classes be sharable between the AST executor and the Bytecode executor?

Or, it's also okay to keep the current simple dictionary and leave some TODO comments here.

codingpotato: We may need to define symbol table classes to handle the global variables in AST component and Bytecode component. But their behavior might be quite different, because the symbol table class in Bytecode component might be used in compile time, instead of runtime. It's hard to decide if they will have a common base class located in SeedLang.Runtime now.
Anyway, I will add a TODO comment to remind us define the symbol table class in the future.

wixette (approved):

PR #48, opened 2021-08-28
Implement block parser

codingpotato:
The block parser can parse text source code of block programs and convert text expression source code to a list of BaseBlock.

### DIFF
--- a/csharp/src/SeedLang.Block/Parser.cs
+++ b/csharp/src/SeedLang.Block/Parser.cs
@@ -1,0 +1,61 @@
+// Copyright 2021 The Aha001 Team.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Collections.Generic;
+using SeedLang.X;
+
+namespace SeedLang.Block {
+  // A parser class which is responsible to convert a block program to an AST tree, an inline
+  // expression text to a list of blocks, and an inline expression text to an AST tree.
+  public class Parser {
+    private class ExressionListener : BlockParser.IExpressionListener {
+      public readonly List<BaseBlock> Blocks = new List<BaseBlock>();
+
+      public void VisitArithmeticOperator(string op) {
+        Blocks.Add(new ArithmeticOperatorBlock { Name = op });
+      }
+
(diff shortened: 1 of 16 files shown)


csharp/src/SeedLang.X/BaseParser.cs
```diff
@@ -0,0 +1,115 @@
...
+using System.Diagnostics;
+using Antlr4.Runtime;
+using Antlr4.Runtime.Tree;
+using SeedLang.Ast;
+using SeedLang.Common;
+
+namespace SeedLang.X {
+  // The abstract base class of the block parser and all SeedX language parsers.
+  //
+  // It provides interfaces to validate the SeedPython source code, and parse it into an AST tree
```
wixette, line 24: "the SeedPython source code" - this class is not specific to SeedPython, isn't it?

csharp/src/SeedLang.X/BlockParser.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Diagnostics;
+using Antlr4.Runtime;
+using Antlr4.Runtime.Tree;
+using SeedLang.Ast;
+using SeedLang.Common;
+
+namespace SeedLang.X {
+  // The text parser of SeedBlock language.
+  //
+  // The BlockParser inherits the interfaces of BaseParser and provides additional interface to
```
wixette, line 24: additional interface -> an additional interface

csharp/src/SeedLang.Block/Parser.cs
```diff
@@ -0,0 +1,61 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Collections.Generic;
+using SeedLang.X;
+
+namespace SeedLang.Block {
+  // A parser class which is responsible to convert a block program to an AST tree, an inline
+  // expression text to a list of blocks, and an inline expression text to an AST tree.
```
wixette, line 20: A clearer expression for the comments:

```
// Converts a block program to an AST tree, an inline expression text to a list of blocks,
// or an inline expression text to an AST tree.
```

(Optional) Is it good to clarify in the comments that SeedLang.Block.Parser is different from Seedlang.X.BlockParser? The two classes look a bit confusing to newcomers. It'd probably be good to comment on the relationship between the two.

codingpotato: I adjusted this comment but still keep the subject "The parser", because this is a comment for a class instead of a method. A statement to mention the diffenece with SeedLang.X.BlockParser is also added.

csharp/src/SeedLang.X/BaseParser.cs
codingpotato: Done.

csharp/src/SeedLang.X/BlockParser.cs
codingpotato: Done.

wixette (approved):

PR #50, opened 2021-09-02
Restructure the organization of SeedLang csharp code

codingpotato:
- Combine all projects into two projects: SeedLang.Shell and SeedLang.
- Remove the Engine class, add a new Executor class in the Runtime component.
- Make most of the classes internal. Only SeedLang.Block, SeedLang.Runtime and SeedLang.Common components have public classes.
- SeedLang.Shell will depends on the interfaces of Runtime to execute the SeedX source code.
- The Unity UI will depends on the interfaces of SeedLang.Block to maintain the block program, and call interfaces of SeedLang.Runtime to execute and visualize block programs.


wixette (approved):
Well done! I love this refactoring!

PR #51, opened 2021-09-04
Run block programs

codingpotato:
* Run text block source code from SeedLang.Shell
* Run block programs from the executor of SeedLang.Runtime

### DIFF
--- a/csharp/src/SeedLang/Runtime/Executor.cs
+++ b/csharp/src/SeedLang/Runtime/Executor.cs
@@ -19,7 +19,7 @@ using SeedLang.Common;
 using SeedLang.X;
 
 namespace SeedLang.Runtime {
-  // An executor class to execute block programs or SeedX source code. The information during
+  // An executor class to execute SeedBlock programs or SeedX source code. The information during
   // execution can be visualized by registered visualizers.
   public class Executor {
     private readonly VisualizerCenter _visualizerCenter = new VisualizerCenter();
@@ -38,28 +38,27 @@ namespace SeedLang.Runtime {
       _visualizerCenter.Unregister(visualizer);
     }
 
-    // Runs a block program.
+    // Runs a SeedBlock program.
     public bool Run(Program program, DiagnosticCollection collection = null) {
       if (program is null) {
         return false;
       }
       DiagnosticCollection localCollection = collection ?? new DiagnosticCollection();
-      foreach (var node in Parser.TryParse(program, localCollection)) {
+      foreach (var node in Converter.TryConvert(program, localCollection)) {
         _executor.Run(node);
       }
       return true;
     }
 
-    // Runs text-based SeedBlock or SeedX source code based on the given programming language and
-    // parse rule.
-    public bool Run(string source, string module, ProgrammingLanguage language, ParseRule rule,
-                    RunType runType, DiagnosticCollection collection = null) {
+    // Runs SeedX source code based on the given SeedX language and run type.
+    public bool Run(string source, string module, SeedXLanguage language, RunType runType,
+                    DiagnosticCollection collection = null) {
       if (string.IsNullOrEmpty(source) || module is null) {
         return false;
       }
       DiagnosticCollection localCollection = collection ?? new DiagnosticCollection();
       BaseParser parser = MakeParser(language);
-      if (parser.TryParse(source, module, rule, localCollection, out AstNode node)) {
+      if (parser.TryParse(source, module, ParseRule.Statement, localCollection, out AstNode node)) {
         switch (runType) {
           case RunType.Ast:
             _executor.Run(node);
@@ -69,11 +68,9 @@ namespace SeedLang.Runtime {
       return false;
     }
 
-    private static BaseParser MakeParser(ProgrammingLanguage language) {
+    private static BaseParser MakeParser(SeedXLanguage language) {
       switch (language) {
-        case ProgrammingLanguage.TextBlock:
-          return new BlockTextParser();
-        case ProgrammingLanguage.Python:
+        case SeedXLanguage.Python:
           return new PythonParser();
         default:
           Debug.Assert(false, $"Not implemented SeedX language: {language}");
--- a/csharp/src/SeedLang/Runtime/Executor.cs
+++ b/csharp/src/SeedLang/Runtime/Executor.cs
@@ -14,6 +14,7 @@
 
 using System.Diagnostics;
 using SeedLang.Ast;
+using SeedLang.Block;
 using SeedLang.Common;
 using SeedLang.X;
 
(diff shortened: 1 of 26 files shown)


csharp/src/SeedLang/Runtime/Executor.cs
```diff
@@ -53,7 +68,7 @@ public class Executor {
       return false;
     }
 
-    private static BaseParser MakeParser(ProgrammingLanguage language) {
+    private static BaseParser MakeXParser(ProgrammingLanguage language) {
```
wixette, line 71: You've changed "MakeParser" to "MakeXParser". Conceptually you intended to use the term X to cover both the SeedBlock language and all the other text-based Seed languages (SeedPython, SeedLua, etc.), right?

Just want to confirm this. If you think we'd better use SeedX as a common name for both the Block language and the text-based languages, I am going to update the main design doc to clarify this.

codingpotato: The intention I changed this name is to make the naming clearer, but I still feel a little confusing. What's your feeling about it?
- Parsers in SeedLang.X covers all text-based Seed languages (**text-based Block language**, SeedPython, SeedLua, etc.). At least we need implement expression parsing of the text-based Block language in SeedLang.X. I also implemented running a text-based Block program from SeedLang.Shell, but it's optional.
- The parser in SeedLang.Block is used to convert among Block language, AST and inline editable text. Is this name confusing for other developer? I cannot figure out a better name for it now.
- They both are used in Executor. This is the reason I changed the name of MakeXParser

wixette: ## For the overall design:

* I personally like the idea that uses `SeedBlock` to represent the unique block language we are defining, while using `SeedX` for all the text-based languages that are derived from a group of existing languages (like Python, Lua, JS, etc.)
* Our block language deserves a unique name `SeedBlock` because it is very different from existing block languages such as Google blockly and Scratch.
* `SeedX` is good for derived languages because `X` here can imply a variable and it can represent the name of any existing programming language.
* If we have to call `SeedBlock` and all other `SeedX` languages together, we can use the term "SeedLang languages". It is clear enough in this context.

## For the parser classes:

Current class structure:

```
SeedLang.X.BaseParser
    SeedLang.X.PythonParser : BaseParser
    SeedLang.X.BlockParser : BaseParser
    SeedLang.X.BlockVisitor
SeedLang.Block.Parser
```

The name BlockParser here is really confusing to newcomers I guess. For two reasons:

* `X` as in `SeedX` is not related to the block language per the overall design.
* `BlockParser` here is not parsing blocks, but parsing the inline text code in a block.

## Idea 1

To remove this confusion, the first and simple idea I've got is to rename SeedLang.X.BlockParser to a rather complex but more expressive name, like `BlockInlineTextParser`. So the class structure will be:

```
SeedLang.X.BaseParser
    SeedLang.X.PythonParser : BaseParser
    SeedLang.X.BlockInlineTextParser : BaseParser
    SeedLang.X.BlockInlineTextVisitor
SeedLang.Block.Parser
```

Here I think it makes sense to keep the class `BlockInlineTextParser` inside the `X` dir since the inline code of SeedBlock is very similar to python code or Lua code and we want to treat them in a consistent way (i.e., parsing them with ANTLR and a consistent wrapper framework).

With this change, the overall design does not need to be updated and the name `SeedX` can still refer to all the text-based languages that are derived from existing languages. Having `BlockInlineTextParser` inside the `X` dir makes this logic not that strict but it's okay and it's not very hard to understand.

## Idea 2

The second idea is to use a generic name like `Parser` to replace the current term `X` in the namespace:

```
SeedLang.Parser.BaseParser
    SeedLang.Parser.PythonParser : BaseParser
    SeedLang.Parser.BlockInlineTextParser : BaseParser
    SeedLang.Parser.BlockInlineTextVisitor
```

## Idea 3

In the case of the above idea, since the term `Parser` is highlighted in the namespace, I think it's not good to still naming the class inside SeedLang.Block as "Parser". We can probably name it as:

```
SeedLang.Block.Converter
```

Or, maybe, separate its functionalities into different util classes, like,

```
SeedLang.Block.InlineTextToBlocks
SeedLang.Block.BlocksToAst
```

Just some random thoughts. Let me know your ideas. Thanks,

codingpotato: Follow idea1 with a little modification:
* Change the name from BlockParser to BlockTextParser, and from BlockVisitior to BlockTextVisitor, because this parser can handle not only inline expression text, but also all the text-based SeedBlock program (include statements).

Summary:
* SeedX only includes Python, Lua, JS, etc.
* Parsers in SeedX component handle text-based SeedBlock and all SeedX source code.
* Parser in SeedBlock component handle real Block program.

wixette: Just to clarify, what is "the text-based SeedBlock program" in your context?

For example, in my mind:

* SeedBlock program is a block-based programming language (instead of a text-based programming language).
* "inline editing" is a feature of SeedBlock. I.e., a number of blocks can be edited by editing plain text content. - but this does not imply that SeedBlock is a text-based language somehow.
* For implementing the "inline editing" feature, we need a text-based parser to convert plain text content to blocks.
* Block eXchange Format (BXF) is a plain text format, but it is only used for serializing and deserializing.

Are you suggesting we can write SeedBlock in plain text and run it with SeedLang.Shell? If so, what format should we use to organize the text of different blocks? Will the format be BXF or something else?

If just considering a convenient testing/integrating process, the SeedLang.Shell or a unit test can still read a BXF text and deserialize BXF to blocks then do the tests, right? It seems we don't need a separate format to represent an entire SeedBlock program.

I guess what we need is just a text grammar that represents all the cases of the texts that can be the target of "inline editing", right? If so, "BlockInlineTextParser" will be clearer than "BlockTextParser".

And, I saw you have defined the language type as `ProgrammingLanguage.TextBlock`. The name `TextBlock` strongly implies that it's a text format of an entire block program. I guess it sounds confusing with the existing format `BXF`...

codingpotato: Because we need define SeedBlock.g4 to generate the ANTL4 block parser to parse the inline expression text, I can also define all the SeedBlock syntax in this grammar file, which can be used to parser the whole text-based block program. It's already supported. Please have a try:
* Checkout to my branch: wip_codingpotato_run_blocks
* Run SeedLang.Shell with following command: dotnet run -l TextBlock
* Input following text-based block program:
```
> set a to 1
> set b to 2
> eval a + b
```
These text-based block source code can be run correctly.

wixette: Interesting!

But I think it will still take a certain amount of effort to improve this grammar definition to make a full-functional text-based programming language since the full feature-set of SeedBlock grammar will be a superset of the necessary grammar set that supports inline editing. There will be a couple of block features that have no need to be inline editable.

Given this, we need to do additional work if we want to define a full "text block" language, even if the amount of the additional work is not that large. I am fine if you think it's not a big deal to keep such a "new" text-based language. But I would suggest we name the "new" language carefully.

`SeedTextBlock` or `SeedBlockText` sounds not good to me due to confusing meanings. At least, we need to distinguish the concept "text code for the inline-editing feature of SeedBlock" from the concept "an entire text representation of SeedBlock" or "a new text-based language derived from SeedBlock". Those two names make users really confusing, e.g., "what the hell is SeedBlock? is it a block-based language? is it a text-based language? or both?"

Probably, can we name the "new" language as:
* `SeedBlockAsText`
* or, being simplified, how about `SeedBAT`
* or, maybe, more advanced, how about `SeedBat`?
* ...

And from the product or marketing point of view, what are the consumer or marketing reasons we need such a new language?
* For serializing and deserializing the block program? Which one is better, BXF or the new text-based language? Obviously, the new text-based language cannot cover all the features of BXF here.
* For users to learn a new text-based language? But why? What's the users' pain point we want to solve?
* For providing 3rd-party developers a debugging interface?
* We can release it since the grammar is already there? - I don't think this is a good reason...

So far I guess the full support of a new text-based language is at a low priority...

wixette: Overall, how about keeping the idea of releasing a "new" text-based language in our mind and:

* Use the names like `BlockInlineTextParser` and `BlockInlineTextVisitor` to only focus on the inline-editing feature for now.
* Name the language type in the executor as `BlockInlineText` for now.
* In the future, go back to this discussion once we are confident that the "new" text-based language is ready. At that time, we can come up with a good name for the "new" language if we think it's reasonable to release it per the perspective of product and marketing.

Your thoughts?

codingpotato: I agree that text-base block program might not be a solid requirement now. I tried to restrict the requirement as following and adjusted the code based on the restriction. I think the code organization will be much clearer. Please have a look again:
- Only implement inline text parsing for block programs and move the parser class to Block namespace.
- SeedLang.Shell only supports running SeedX programs interactively. (SeedBlock programs can only be run by Shell by loading the BXF file in the future). So the language definition only need cover SeedX languages.

wixette (approved):

csharp/src/SeedLang/Runtime/Executor.cs
wixette: Great! I agree with your decisions.

PR #54, opened 2021-09-21
Add NuGet package metadata

wixette:
Add metadata so that SeedLang can be published to NuGet manually. Will add a versioning and publishing script to aotumate the process.

See the published versions at https://www.nuget.org/packages/SeedLang/

It sounds like a promising way to publish SeedLang to NuGet first then import the package to Unity (via a third-party extension, https://github.com/GlitchEnzo/NuGetForUnity). Although, there is still an unsolved dependency - System.Text.Json is failed to run in Unity - I will do more tests and try to fix it (e.g. by reverting back to Newtonsoft.Json) shortly.


codingpotato (approved):
LGTM

PR #55, opened 2021-09-23
Revert back from System.Text.Json to Newtonsoft.Json

wixette:
Unity has no official support of .Net 5.0 for now. Although it's possible to import NuGet package System.Text.Json for netstandard 2.0, the dependencies are very complex and hard to maintain. We have to manually choose and import 7 DLLs and disable "Validate References" for them in Unity.

With the solution to depend on Newtonsoft.Json, it's easy to import SeedLang.dll to Unity project without importing Newtonsoft.Json since Unity already provides one by default.

Please take a look at the experimental branch https://github.com/aha-001/SeedCalc/tree/wip_wixette_integration to see how to import SeedLang.dll into a Unity project and run a SeedLang executor in Unity scripts.

There are behavior differences between System.Text.Json and Newtonsoft.Json, especially in the strict deserialization features. And it needs non-trivial effort to keep the same behavior with Newtonsoft.Json. But I think it's okay to stay with the default behavior of Newtonsoft.Json considering the needs of SeedLang.

### DIFF
--- a/csharp/src/SeedLang/Block/BxfConstants.cs
+++ b/csharp/src/SeedLang/Block/BxfConstants.cs
@@ -12,9 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System.Text.Json;
-using System.Text.Json.Serialization;
-
 namespace SeedLang.Block {
   // Common settings used by the Block Exchange Format.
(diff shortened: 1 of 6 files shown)


codingpotato (approved):
LGTM

PR #57, opened 2021-09-23
Improve the Converter interface

wixette:
* Supports Validate and Parse separately
* Parses inline text strings with invalid tokens
* Reports error info via Diagnostic collection
* Remove unused code in unittests

### DIFF
--- a/csharp/src/SeedLang/Block/Converter.cs
+++ b/csharp/src/SeedLang/Block/Converter.cs
@@ -23,44 +23,75 @@ namespace SeedLang.Block {
   public class Converter {
     private class InlineTextListener : InlineTextParser.IInlineTextListener {
       public readonly List<BaseBlock> Blocks = new List<BaseBlock>();
+      public IList<TextRange> InvalidTokenRanges { get; private set; }
 
-      public void VisitArithmeticOperator(string op) {
-        Blocks.Add(new ArithmeticOperatorBlock { Name = op });
+      public InlineTextListener(IList<TextRange> invalidTokenRanges) {
+        InvalidTokenRanges = invalidTokenRanges;
       }
 
-      public void VisitCloseParen() {
-        Blocks.Add(new ParenthesisBlock(ParenthesisBlock.Type.Right));
+      public void VisitArithmeticOperator(string op, TextRange range) {
+        Blocks.Add(new ArithmeticOperatorBlock { Name = op, InlineTextReference = range });
       }
 
-      public void VisitIdentifier(string name) {
+      public void VisitCloseParen(TextRange range) {
+        Blocks.Add(new ParenthesisBlock(ParenthesisBlock.Type.Right) {
+          InlineTextReference = range
+        });
+      }
+
+      public void VisitIdentifier(string name, TextRange range) {
         throw new System.NotImplementedException();
       }
 
-      public void VisitNumber(string number) {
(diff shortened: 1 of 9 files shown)


csharp/src/SeedLang/Block/Converter.cs
```diff
@@ -23,44 +23,75 @@ namespace SeedLang.Block {
...
+
+      public void VisitInvalidToken(TextRange range) {
+        InvalidTokenRanges?.Add(range);
+      }
     }
 
-    // Converts an expression inline text to a list of blocks.
-    public static IEnumerable<BaseBlock> InlineTextToBlocks(string text) {
-      Debug.Assert(!string.IsNullOrEmpty(text), "Inline text shall not be null or empty.");
+    // Checks if an inline text is a valid expression. Detailed info will be stored in collection if
```
codingpotato, line 65: Detailed info -> Detailed diagnostic info.

csharp/src/SeedLang/Block/InlineTextParser.cs
```diff
@@ -28,62 +29,72 @@ internal class InlineTextParser : BaseParser {
...
+            listener.VisitArithmeticOperator(token.Text, TextRange.RangeOfToken(token));
             break;
           case SeedBlockLexer.SUB:
             if (lastTokenType == SeedBlockParser.NUMBER ||
                 lastTokenType == SeedBlockParser.CLOSE_PAREN) {
-              listener.VisitArithmeticOperator(token.Text);
+              listener.VisitArithmeticOperator(token.Text, TextRange.RangeOfToken(token));
             } else {
               negative = true;
+              negativeToken = token;
```
codingpotato, line 71: negative and nagativeToken are duplicated. Is it better to remove negative variable and check if negativeToken is not null to see if met a negative sign?

csharp/src/SeedLang/Common/TextRange.cs
```diff
@@ -50,5 +51,16 @@ public class TextRange : Range {
...
     }
+
+    public static TextRange RangeOfToken(IToken token) {
+      return RangeOfTokens(token, token);
+    }
+
+    public static TextRange RangeOfTokens(IToken start, IToken end) {
+      // TODO: need scan the source string to calculate the end column if the end token is in
+      // multiple lines.
+      return new TextRange(start.Line, start.Column,
```
codingpotato, line 62: I feel it's not good that the TextRange class in Common component depends on Antlr4.Runtime. Maybe we can limit this kind of dependency in X and Block components. It's better to keep the Token to TextRange conversion in SeeLang.X or SeedLang.Block.

codingpotato (changes requested):

csharp/src/SeedLang/Block/Converter.cs
wixette: Done

csharp/src/SeedLang/Block/InlineTextParser.cs
wixette: Done

csharp/src/SeedLang/Common/TextRange.cs
wixette: Done. Moved the two methods to a utility class in SeedLang.X.

codingpotato (approved):
LGTM

PR #68, opened 2021-09-27
Fixes #67

### LINKED ISSUE #67: InlineTextParser.VisitInlineText doesn't return "-" when "-" is at the tail of the text and the expression is incomplete
E.g., InlineTextParser.VisitInlineText returns nothing when the input is "-", "3-", or "3+4-".

wixette:
Fix the bug: Negative is not returned when it is the last token of an incomplete expression.


codingpotato (approved):
LGTM

PR #74, opened 2021-10-13
NuGet package versioning and build/release guide

wixette:
With this change, we can manually publish NuGet packages with a
semantic versioning mechanism.

I think a manual process is good enough for the current situation.
We can automate the release pipeline when it's too complex
to do it manually.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -4,11 +4,11 @@
 
 ### The NuGet package of SeedLang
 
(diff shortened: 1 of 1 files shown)


codingpotato:
One question:
When we release a version on NuGet, do we need also apply a tag on Github, so that we can rebuild this version later to reproduce or debug issues?

codingpotato (approved):
LGTM

PR #78, opened 2021-10-21
Implement the virtual machine to execute binary expressions

codingpotato:
- Run `dotnet run -p src/SeedLang.Shell -t Bytecode` from `csharp` folder
   Only support binary expressions now.
   TODO: Variables and unary expressions will be implemented in future PRs.

- Add a benchmark folder for performance benchmarks.
   Run `dotnet run -c Release -p benchmark/SeedLang.Benchmark`

- Remove IValue interface, and move Value classes from Ast to Runtime component.

### DIFF
--- a/csharp/src/SeedLang/Interpreter/Chunk.cs
+++ b/csharp/src/SeedLang/Interpreter/Chunk.cs
@@ -45,7 +45,8 @@ namespace SeedLang.Interpreter {
       var sb = new StringBuilder();
       for (int i = 0; i < _bytecode.Count; ++i) {
         sb.Append($"{_bytecode[i],-20}{ConstantOperandToString(_bytecode[i])}");
-        sb.AppendLine(i < _ranges.Count && _ranges[i] is Range range ? $"{range}" : "");
+        Debug.Assert(i < _ranges.Count);
+        sb.AppendLine(_ranges[i] is null ? "" : $"{_ranges[i]}");
       }
       return sb.ToString();
     }


csharp/src/SeedLang/Interpreter/Chunk.cs
```diff
@@ -15,48 +15,68 @@
...
+
+    private readonly List<Range> _ranges = new List<Range>();
 
     public override string ToString() {
       var sb = new StringBuilder();
-      foreach (var instr in _bytecode) {
-        sb.AppendLine($"{instr,-20}{ConstantOperandToString(instr)}");
+      for (int i = 0; i < _bytecode.Count; ++i) {
+        sb.Append($"{_bytecode[i],-20}{ConstantOperandToString(_bytecode[i])}");
+        sb.AppendLine(i < _ranges.Count && _ranges[i] is Range range ? $"{range}" : "");
```
wixette, line 48: If the code logic should ensure that "The length of Bytecode and Range lists shall be the same", it'd be better to guard "I < _ranges.Count" with an Assert in debug mode, right? Does "_ranges[i] is Range" fall in the same situation?

csharp/src/SeedLang/Interpreter/VM.cs
```diff
@@ -0,0 +1,93 @@
...
+    private VMValue[] _registers;
+
+    internal VM(VisualizerCenter visualizerCenter = null) {
+      _visualizerCenter = visualizerCenter ?? new VisualizerCenter();
+    }
+
+    internal void Run(Chunk chunk) {
+      _chunk = chunk;
+      _registers = new VMValue[chunk.RegisterCount];
+      int pc = 0;
```
wixette, line 32: For the data type of PC, is `uint` better than `int`? I don't have a preference here. Just want to confirm that you have already considered all the corner cases. E.g. when a `JMP sBx` instruction results in a negative PC?

csharp/benchmark/SeedLang.Benchmark/Program.cs
```diff
@@ -12,19 +12,12 @@
...
-  // There are different kinds of value types in the AST and interpreter components. They all
-  // implement this value interface.
-  public interface IValue {
-    ValueType Type { get; }
-    double ToNumber();
-    string ToString();
+namespace SeedLang.Benchmark {
+  class Program {
+    static void Main(string[] _) {
+      BenchmarkRunner.Run<BinaryExpressionBenchmark>();
```
wixette, line 20: Suggestion: we have a benchmark program now. It's great to track the benchmark result every time we improve the codebase. Other than a decent automatic tracking system (e.g., run the benchmark in CI and record the result somewhere automatically), we can at least create a separate GitHub issue to manually log those results - simply comment on the issue each time we have made a non-trivial code change.

Of course, my suggestion doesn't block this pull request at all.

PR #84, opened 2021-11-10
Refactor values

codingpotato:
- Add null and boolean value types.
- Add IValue interface for runtime value and VMValue to implement.
- Optimize VMValue to readonly struct, and use `in` and `ref readonly` return to avoid copying.


csharp/src/SeedLang/Runtime/ValueHelper.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    internal static bool NumberToBoolean(double value) {
+      return value != 0;
+    }
+
+    internal static string NumberToString(double value) {
+      return $"{value}";
+    }
+
+    internal static bool StringToBoolean(string value) {
```
wixette, line 64: Why isn't this string-to-bool logic consistent with Python's implicit conversion from string to bool?

For example, see http://anh.cs.luc.edu/handsonPythonTutorial/boolean.html: 

```python
>>> bool(None)
False
>>> bool('')
False
>>> bool('0')
True
>>> bool('False')
True
```

codingpotato: Fixed

wixette (approved):

PR #86, opened 2021-11-15
Add compare expressions and implement AST execution

codingpotato:
Compare expressions with AST execution is implemented, like:
`1 < 2 < 3 == 3`

Bytecode execution hasn't implemented.

### DIFF
--- a/csharp/src/SeedLang.Shell/Repl.cs
+++ b/csharp/src/SeedLang.Shell/Repl.cs
@@ -23,7 +23,7 @@ namespace SeedLang.Shell {
   internal sealed class Repl {
     private class Visualizer : IVisualizer<AssignmentEvent>,
                                IVisualizer<BinaryEvent>,
-                               IVisualizer<CompareEvent>,
+                               IVisualizer<ComparisonEvent>,
                                IVisualizer<EvalEvent> {
       // Current executed source code.
       public string Source { get; set; }
@@ -39,14 +39,14 @@ namespace SeedLang.Shell {
             {BinaryOperator.Modulus, "%"},
           };
 
-      private readonly Dictionary<CompareOperator, string> _compareOperatorStrings =
-          new Dictionary<CompareOperator, string>() {
-            {CompareOperator.Less, "<"},
-            {CompareOperator.Great, ">"},
-            {CompareOperator.LessEqual, "<="},
-            {CompareOperator.GreatEqual, ">="},
-            {CompareOperator.EqualEqual, "=="},
-            {CompareOperator.NotEqual, "!="},
+      private readonly Dictionary<ComparisonOperator, string> _comparisonOperatorStrings =
+          new Dictionary<ComparisonOperator, string>() {
+            {ComparisonOperator.Less, "<"},
+            {ComparisonOperator.Greater, ">"},
+            {ComparisonOperator.LessEqual, "<="},
+            {ComparisonOperator.GreaterEqual, ">="},
+            {ComparisonOperator.EqEqual, "=="},
+            {ComparisonOperator.NotEqual, "!="},
           };
 
       public void On(AssignmentEvent ae) {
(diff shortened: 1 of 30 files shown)


csharp/src/SeedLang.Shell/Repl.cs
```diff
@@ -23,20 +23,31 @@ namespace SeedLang.Shell {
...
+            {BinaryOperator.Divide, "/"},
+            {BinaryOperator.FloorDivide, "//"},
+            {BinaryOperator.Power, "**"},
+            {BinaryOperator.Modulus, "%"},
+          };
+
+      private readonly Dictionary<CompareOperator, string> _compareOperatorStrings =
+          new Dictionary<CompareOperator, string>() {
+            {CompareOperator.Less, "<"},
+            {CompareOperator.Great, ">"},
```
wixette, line 45: The name should be `Greater`, or the full name `GreaterThan`.

I think we can make the in-code names consistent with Python's in-code token names. See the token name list of Python at:

https://github.com/python/cpython/blob/main/Parser/token.c

csharp/src/SeedLang.Shell/Repl.cs
```diff
@@ -23,20 +23,31 @@ namespace SeedLang.Shell {
...
+            {CompareOperator.LessEqual, "<="},
+            {CompareOperator.GreatEqual, ">="},
+            {CompareOperator.EqualEqual, "=="},
```
wixette, line 48: `EqEqual`? to be consistent with https://github.com/python/cpython/blob/main/Parser/token.c ?

csharp/src/SeedLang/Ast/Executor.cs
```diff
@@ -73,6 +73,57 @@ internal sealed class Executor : AstWalker {
...
+            throw new System.NotImplementedException(
+                $"Unsupported compare operator: {compare.Ops[i]}");
+        }
+        if (!currentResult) {
+          break;
+        }
+      }
+      _expressionResult = new BooleanValue(currentResult);
+      if (!_visualizerCenter.ComparePublisher.IsEmpty()) {
+        // Evaluate short circuit expressions for visualizers notification.
```
wixette, line 114:
```suggestion
        // Evaluates short circuit expressions for visualizers notification.
```

csharp/src/SeedLang/Ast/Expressions.cs
```diff
@@ -24,6 +25,12 @@ internal abstract class Expression : AstNode {
       return new BinaryExpression(left, op, right, range);
     }
 
+    // The factory method to create a compare expression.
+    internal static CompareExpression Compare(Expression first, CompareOperator[] ops,
```
wixette, line 29: Naming: `ComparisonExpression`

csharp/src/SeedLang/Runtime/Events.cs
```diff
@@ -12,24 +12,49 @@
...
+        base(range) {
       Left = left;
       Op = op;
       Right = right;
       Result = result;
-      Range = range;
+    }
+  }
+
+  // An event which is triggered when a compare expression is evaluated.
```
wixette, line 44: Naming: `Comparison expression`

Here and others

csharp/src/SeedLang/Ast/Executor.cs
codingpotato: I have another idea about this event. These short circuit expressions shall not be evaluated, so that users can be indicated that some expressons are short circuited, and not evaluated.

codingpotato: Please have a review again for this.

csharp/src/SeedLang/Ast/Expressions.cs
codingpotato: Done

csharp/src/SeedLang/Runtime/Events.cs
codingpotato: Done

csharp/src/SeedLang.Shell/Repl.cs
codingpotato: Done

csharp/src/SeedLang.Shell/Repl.cs
codingpotato: Done

wixette (approved):

PR #87, opened 2021-11-20
Refactor SeedPython grammar

codingpotato:
- Refactor SeedPython grammars based on Python 3.10
- Reimplement SeedBlockInlineText and SeedPython visitors.


grammars/Common.g4
```diff
@@ -12,66 +12,91 @@
...
-  | expression addSubOperator expression        # addSubExpression
-  | expression (comparisonOperator expression)+ # comaprisonExpression
-  | IDENTIFIER                                  # identifier
-  | NUMBER                                      # number
-  | OPEN_PAREN expression CLOSE_PAREN           # grouping;
-
-comparison:
-  expression (comparisonOperator expression)+;
+star_expressions:
+  star_expression (COMMA star_expression)+ COMMA? # mltiple_star_expressions
```
wixette, line 36:
```suggestion
  star_expression (COMMA star_expression)+ COMMA? # multiple_star_expressions
```

grammars/SeedPython.g4
```diff
@@ -13,75 +13,58 @@
...
+program: statements? EOF;
+
+statements: statement+;
+statement: compound_stmt | simple_stmts;
+simple_stmts:
+  simple_stmt (SEMICOLON simple_stmt)+ SEMICOLON? # multiple_simple_stmts
+  | simple_stmt                                   # single_simple_stmt;
+simple_stmt:
+  assignment         # assignment_placeholder
+  | star_expressions # expression_stmt
```
wixette, line 33: Why do you define `expression_stmt` as `star_expressions`?

In https://github.com/bkiers/python3-parser/blob/master/src/main/antlr4/nl/bigo/pythonparser/Python3.g4 the name `star_expr` is defined as:

```
/// star_expr: ['*'] expr
star_expr
 : '*'? expr
 ;
```

But so far it's defined as below in our grammar:

```
star_expressions:
  star_expression (COMMA star_expression)+ COMMA? # mltiple_star_expressions
  | star_expression 
star_expression: expression;
```

I didn't find anything like `'*'? expr` in the grammar. It's not reasonable to still call it as "star_something".

Will our grammar become the same as the ANTLR's origin grammar a bit later? 

Since our grammar hasn't covered the real star-expr case like `a, *b = 1, 3`, or `def foo(a, *b)`, can we avoid using a confusing name for now?

grammars/Common.g4
codingpotato: Done

grammars/SeedPython.g4
codingpotato: Agree that the star-expr will not be our requirement in very long time. So I remove all rules related to the star-expr to make it clear.

wixette (approved):

PR #92, opened 2021-11-26
Add newline to separate simple statements

### LINKED ISSUE #90: Eval with parenthesis fails
In SeedLang.Shell, Input:

```
6(5*6)
```

Expected result: reporting syntax errors.

Actual result (no error):

```
---------- Source ----------
6(5*6)

---------- Run ----------
6(5*6): 5 * 6 = 30
6(5*6): Eval result: 30
```

Similar with the cases like:

```
(234+3432)3
```

No error is reported:

```
---------- Source ----------
(234+3432)3

---------- Run ----------
(234+3432)3: Eval result: 3
```

codingpotato:
Fix #90

### DIFF
--- a/csharp/src/SeedLang.Shell/Repl.cs
+++ b/csharp/src/SeedLang.Shell/Repl.cs
@@ -119,7 +119,8 @@ namespace SeedLang.Shell {
         WriteSource(visualizer.Source, syntaxTokens);
         Console.WriteLine("---------- Run ----------");
         var runCollection = new DiagnosticCollection();
-        executor.Run(visualizer.Source, "", _language, _runType, runCollection);
+        string source = visualizer.Source + Environment.NewLine;
+        executor.Run(source, "", _language, _runType, runCollection);
         foreach (var diagnostic in runCollection.Diagnostics) {
           if (diagnostic.Range is TextRange range) {
             visualizer.WriteSourceWithHighlight(range);
--- a/csharp/src/SeedLang/X/PythonVisitor.cs
+++ b/csharp/src/SeedLang/X/PythonVisitor.cs
@@ -42,6 +42,11 @@ namespace SeedLang.X {
       return Visit(statements);
     }
 
+    public override AstNode VisitSingle_simple_stmt(
+        [NotNull] SeedPythonParser.Single_simple_stmtContext context) {
+      return Visit(context.simple_stmt());
+    }
+
     public override AstNode VisitExpression_stmt(
         [NotNull] SeedPythonParser.Expression_stmtContext context) {
       return VisitorHelper.BuildExpressionStatement(context.expressions(), this);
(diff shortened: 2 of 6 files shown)


csharp/src/SeedLang.Shell/Repl.cs
```diff
@@ -119,7 +119,8 @@ private class Visualizer : IVisualizer<AssignmentEvent>,
         WriteSource(visualizer.Source, syntaxTokens);
         Console.WriteLine("---------- Run ----------");
         var runCollection = new DiagnosticCollection();
-        executor.Run(visualizer.Source, "", _language, _runType, runCollection);
+        string source = visualizer.Source + Environment.NewLine;
```
wixette, line 122: Just to confirm: does this change require every source code that is to be executed by SeedLang has to end with "\n"? For the case when SeedLang.Shell supports a feature to run a SeedPython file directly:

```
SeedLang.Shell.exe ... -c input.py
```
And user forgets to add the trailing "\n" to the input.py file. Will SeedLang.Shell adds a trailing NewLine to the input file automatically? Or, has the current grammar definition already covered the <EOF> issue when parsing simple statements?

codingpotato: According to the definition of Python3 grammar:
```
file: [statements] ENDMARKER 
interactive: statement_newline
statements: statement+ 
statement: compound_stmt  | simple_stmts 
statement_newline:
    | compound_stmt NEWLINE 
    | simple_stmts
    | NEWLINE 
    | ENDMARKER 
simple_stmts:
    | simple_stmt !';' NEWLINE  # Not needed, there for speedup
    | ';'.simple_stmt+ [';'] NEWLINE
```
Python uses `file` and `interactive` rules to distinguish these two different use cases. I think it might not be needed now, one `program` rule is enough to cover these two. I need verify it when the file running case is implemented.
And from above grammar definition, it seems `NEWLINE` is needed for `simple_stmt`, but not needed for `compound_stmt`. I need to verify it as well.
Let's keep it open, and finalize it when the file running case is implemented. Anyway the solution could be adding NEWLINE at the end of the file if users miss it.

wixette: If it's the case "anyway the solution could be adding NEWLINE at the end of the file if users miss it", why don't we check and add the missing NEWLINE inside the logic of executor.Run() for SeedPython?

codingpotato: There is one problem to add NEWLINE inside SeedLang library, because ANTLR will report an error range if there is errors happening, and sometimes the NEWLINE is included in the error range. SeedLang.Shell can not handle the error range correctly in this case if the source is modified by SeedLang library.

wixette: OK. I see. Let's move forward and try to solve this in a better way later.

wixette (approved):

PR #104, opened 2021-12-11
Implement a list value type

codingpotato:
- Refector value types.
  I think we don't need two kinds of value types for AST and VM. And it's a little difficult to keep both the inheritance structure and efficiency when I try to add the list type. So the inheritance value types are removed, and only keep the struct value type.
  A ValueWrapper is also added to wrap the value object, implement the IValue interface. There are some changes for the IValue interface as well.

- Implement a list value type.


csharp/src/SeedLang/Runtime/ValueHelper.cs
```diff
@@ -16,72 +16,72 @@
...
+    internal static double Power(in Value lhs, in Value rhs) {
+      double result = System.Math.Pow(lhs.AsNumber(), rhs.AsNumber());
       CheckOverflow(result);
       return result;
     }
 
-    internal static double Modulo<Value>(in Value lhs, in Value rhs) where Value : IValue {
-      double result = lhs.Number % rhs.Number;
+    internal static double Modulo(in Value lhs, in Value rhs) {
+      double result = lhs.AsNumber() % rhs.AsNumber();
```
wixette, line 66: Should `rhs` be also checked on `DivideByZero` here?

csharp/src/SeedLang/Runtime/Value.cs
```diff
@@ -0,0 +1,220 @@
...
+      None,
+      Boolean,
+      Number,
+      String,
+      List,
+    }
+
+    private readonly ValueType _type;
+    private readonly double _number;
+    private readonly BoxValue _box;
```
wixette, line 37: Per the actual usage of `_box` in the `Value` class, I guess a better name for `_box` could be `_valueOfReferenceType` or `_objOfReferenceType` or even `_nullableValue`. I personally like `_nullableValue`. See https://docs.microsoft.com/en-us/dotnet/api/system.nullable-1?view=net-6.0, A type is said to be nullable if it can be assigned a value or can be assigned null, which means the type has no value whatsoever. By default, all reference types, such as String, are nullable, but all value types, such as Int32, are not.

csharp/src/SeedLang/Runtime/ValueHelper.cs
codingpotato: Forgot it. Graeat catch. Will add it.

csharp/src/SeedLang/Runtime/BoxValue.cs
```diff
@@ -0,0 +1,84 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+
+namespace SeedLang.Runtime {
+  internal class BoxValue {
```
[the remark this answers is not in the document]
codingpotato: My intention is to separate values on the stack (none, boolean and number), and ones on the heap (string, list, table, function, etc.) We may have our own GC implementation for the heap objects.
It seems box is not a good name to express this intention. So I changed it to `HeapObject`. What's your thought about it?

csharp/src/SeedLang/Runtime/Value.cs
codingpotato: Change to `HeapObject`

csharp/src/SeedLang/Runtime/BoxValue.cs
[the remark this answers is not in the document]
wixette: HeapObject sounds good to me!

csharp/src/SeedLang/Runtime/HeapObject.cs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    internal string AsString() {
+      switch (_object) {
+        case string str:
+          return str;
+        case List<Value> list:
+          return list.ToString();
+        default:
+          throw new NotImplementedException($"Unsupported box type: {_object}");
```
wixette, line 45: "box type", or "heap object type"?

Here and the rest.

csharp/src/SeedLang/Runtime/Value.cs
```diff
@@ -0,0 +1,220 @@
...
+
+    private Value(double value) {
+      _type = ValueType.Number;
+      _number = value;
+      _object = null;
+    }
+
+    private Value(string value) : this(ValueType.String, HeapObject.String(value)) { }
+
+    private Value(ValueType type, HeapObject box) {
```
wixette, line 53: box -> obj or heapObj

Here and the rest.

codingpotato: Done

csharp/src/SeedLang/Runtime/HeapObject.cs
codingpotato: Done

wixette (approved):

PR #105, opened 2021-12-14
Parse and execute list and subscript expressions

codingpotato:
Following bubble sort code can be run:
```python
array = [64, 34, 25, 12, 22, 11, 90]
n = 7
i = 0
while i < n:
  j = 0
  while j < n - i - 1:
    if array[j] > array[j + 1]:
      temp = array[j]
      array[j] = array[j + 1]
      array[j + 1] = temp
    j = j + 1
  i = i + 1
array
```

TODO: there might be some exceptions that are not handled.

### DIFF
--- a/csharp/src/SeedLang/Ast/AstStringBuilder.cs
+++ b/csharp/src/SeedLang/Ast/AstStringBuilder.cs
@@ -166,6 +166,21 @@ namespace SeedLang.Ast {
       Exit();
     }
 
+    protected override void Visit(ListExpression list) {
+      Enter(list);
+      foreach (Expression expr in list.Exprs) {
+        Visit(expr);
+      }
+      Exit();
+    }
+
+    protected override void Visit(SubscriptExpression subscript) {
+      Enter(subscript);
+      Visit(subscript.Expr);
+      Visit(subscript.Index);
+      Exit();
+    }
+
     protected override void Visit(AssignmentStatement assignment) {
       Enter(assignment);
       Visit(assignment.Identifier);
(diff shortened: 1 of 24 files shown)


wixette (commented):
I haven't dug into the code deeply. Here are some quick feedback per my ad-hoc test with your pr:

1) empty list is not supported yet

```shell
>>> a = []

Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at SeedLang.X.SeedPythonVisitor.VisitList(ListContext context) in /Users/ygwang/src/SeedLang/csharp/src/SeedLang/X/SeedPythonVisitor.cs:line 231
```

2) apply the index `[]` operator to a non-list value results in an exception:

```shell
>>> a = 3
>>> a[0]
>>> a = [3]
>>> a[0][0]

---------- Run ----------
Unhandled exception. System.NotImplementedException: The method or operation is not implemented.
   at SeedLang.Runtime.Value.get_Item(Int32 index) in /Users/ygwang/src/SeedLang/csharp/src/SeedLang/Runtime/Value.cs:line 145
```

3) Non-integer index is not detected as out-of-range nor type-error:

```shell
>>> a = [1, 2]
>>> a[0.1]

Eval result: 1
```

While in Python3:

```shell
>>> a = [1, 2]
>>> a[0.1]
TypeError: list indices must be integers or slices, not float
```

4) Reference cycles are not detected and handled when printing the evaluation result:

In Python3:

```shell
>>> a = [1, 2]
>>> b = [a]
>>> a [1] = b
>>> a
[1, [[...]]]
>>> b
[[1, [...]]]
```

In this PR of SeedLang:

```shell
>>> a = [1, 2]
>>> b = [a]
>>> a[1] = b
>>> a

Stackoverflow
...
```

See https://devguide.python.org/garbage_collector/#identifying-reference-cycles for Python's GC design doc related to reference cycles.

codingpotato:
- Item 1: Fixed in 9e36c7d
- Item 2: Fixed in aca9d79
- Item 3: Fixed in 4ec9317
- Item4: I need think about how to handle it. Added an issue #106 to track it.

wixette (approved):

PR #110, opened 2021-12-23
Support function definition, call functions and return from functions

codingpotato:
Following code is supported:
```python
def factorial(n):
  if n == 1:
    return 1
  else:
    return n * factorial(n - 1)
factorial(10)
```

### DIFF
--- a/csharp/src/SeedLang/Ast/Environment.cs
+++ b/csharp/src/SeedLang/Ast/Environment.cs
@@ -38,7 +38,7 @@ namespace SeedLang.Ast {
       }
     }
 
-    private readonly LinkedList<Frame> _frames = new LinkedList<Frame>();
+    private readonly Stack<Frame> _frames = new Stack<Frame>();
 
     internal Environment() {
       // Enters the first (global) scope.
@@ -46,11 +46,11 @@ namespace SeedLang.Ast {
     }
 
     internal void EnterScope() {
-      _frames.AddFirst(new Frame());
+      _frames.Push(new Frame());
     }
 
     internal void ExitScope() {
-      _frames.RemoveFirst();
+      _frames.Pop();
     }
 
     internal bool ContainsVariable(string name) {
@@ -73,7 +73,7 @@ namespace SeedLang.Ast {
     }
 
     internal void SetVariable(string name, in Value value) {
-      _frames.First.Value.SetVariable(name, value);
+      _frames.Peek().SetVariable(name, value);
     }
   }
 }
--- a/csharp/src/SeedLang/Ast/Environment.cs
+++ b/csharp/src/SeedLang/Ast/Environment.cs
@@ -1,0 +1,77 @@
+// Copyright 2021 The Aha001 Team.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using SeedLang.Common;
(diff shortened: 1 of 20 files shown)


csharp/src/SeedLang/Ast/Environment.cs
```diff
@@ -0,0 +1,79 @@
...
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using SeedLang.Common;
+using SeedLang.Runtime;
+
+namespace SeedLang.Ast {
+  // The environment class to store scoped variables. A new frame is inserted at the first of a
+  // linked list when a new scope is entered.
+  internal class Environment {
```
wixette, line 23: The name "environment" sounds a bit generic to me. In many cases, "environment" means the whole global environment of the runtime. Doesn't a common name like "CallStack" work here? I think the name "Call Stack" is more common when talking about the structure that holds a couple of function call frames. For example, C# has a diagnostic class named "StackFrame":

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stackframe?view=net-6.0

In its description: Provides information about a StackFrame, which represents a function call on the **call stack** for the current thread.

I am not that sure about this naming issue since you are using this class more like a stacked variable pool (GetVariable/SetVariable)... Just an open discussion...

csharp/src/SeedLang/Ast/Environment.cs
```diff
@@ -0,0 +1,79 @@
...
+        Debug.Assert(_variables.ContainsKey(name));
+        return _variables[name];
+      }
+
+      internal void SetVariable(string name, in Value value) {
+        _variables[name] = value;
+      }
+    }
+
+    private readonly LinkedList<Frame> _frames = new LinkedList<Frame>();
```
wixette, line 41: Which data structure is better, C# Stack class or LinkedList class?

csharp/src/SeedLang/Ast/Executor.cs
```diff
@@ -39,6 +43,26 @@ internal sealed class Executor : AstWalker {
...
 
+    // Calls a AST function with given arguments.
+    internal Value Call(FunctionStatement func, Value[] arguments) {
+      if (func.Parameters.Length != arguments.Length) {
+        throw new DiagnosticException(SystemReporters.SeedAst, Severity.Fatal, "", null,
+                                      Message.RuntimeErrorIncorrectArgsCount);
+      }
+      _env.EnterScope();
+      for (int i = 0; i < func.Parameters.Length; i++) {
+        _env.SetVariable(func.Parameters[i], arguments[i]);
```
wixette, line 54: Here is a question not quite related to this pr: have you considered the future improvement to support closures and coroutines? For example, Lua has an "upvalue" design to support closures and co-routines...

I think we just need to make sure the current design is extendable to be upgraded to a closure-friendly design in the future.

csharp/src/SeedLang/Ast/Environment.cs
codingpotato: The "CallStack" will be good for a class to handle the call stack. But this class will exceed this responsibility. It only holds call frames now, but we need definitly support block-level variable scopes in the future. That is a new scope will be entered when a block is start to execute, and it will be exited when the block is ended. Then it's not just "CallStack".
Do you have any other idea about the naming of this class?

csharp/src/SeedLang/Ast/Environment.cs
codingpotato: I had concern about the performance of iterating all Stack elements when implemented this class. But it should not be a problem when I re-think about it. So changed to Stack.

csharp/src/SeedLang/Ast/Executor.cs
codingpotato: I have plan to implement "upvalue". It's not too hard, just need add a Upvalue value type, and store the outside stack values into it in the returned closure. We can implement it in the future.

csharp/src/SeedLang/Ast/Environment.cs
wixette: No better name for now. Let keep the name "Environment".

csharp/src/SeedLang/Ast/Executor.cs
wixette: Great!

wixette (approved):

PR #117, opened 2022-01-01
Parse and execute "while" statements in VM

codingpotato:


csharp/src/SeedLang/Interpreter/Compiler.cs
```diff
@@ -96,12 +78,64 @@ private struct ExpressionInfo {
     }
 
     protected override void Visit(ComparisonExpression comparison) {
-      throw new NotImplementedException();
+      // TODO: support multiple operators. Current implementation only supports two operators.
```
wixette, line 81: What are the "two operators" that are currently supported? Should they actually be "registers" to hold values? What is the case that requires "multiple operators"? Can you show an example in the comments?

csharp/src/SeedLang/Interpreter/Compiler.cs
```diff
@@ -96,12 +78,64 @@ private struct ExpressionInfo {
...
+          expectedResult = true;
+          break;
+        case ComparisonOperator.NotEqual:
+          op = Opcode.EQ;
+          expectedResult = false;
+          break;
+      }
+      _chunk.Emit(op, expectedResult ? 1u : 0u, first, second, comparison.Range);
+      if (needFirstRegister) {
+        _registerAllocator.DeallocateVariable();
```
wixette, line 124: There are many `if (condition) DeallocateVariable()` code blocks. Could there be resource leak risks if those kind of conditions are not handled carefully? ... Just to make sure that you have already considered different solutions and still think this "allocate-deallocate" pair is the best fit.

csharp/src/SeedLang/Interpreter/Compiler.cs
codingpotato: Modified the comments and added examples.
There was typos in the comments. It shall be `multiple operands` and `two operands`.

csharp/src/SeedLang/Interpreter/Compiler.cs
codingpotato: Added expression scope concept in RegisterAllocator to simplify client logic. Allocated temporary variables will be deallocated automatically when exiting expression scopes.

wixette: This is cool!

wixette (approved):
LGTM

PR #120, opened 2022-01-09
Implement function define, call, and return statements in SeedVM

codingpotato:
Following code can be run by SeedVM:
```python
def f():
    sum = 0
    i = 1
    while i <= 10000000:
        sum = sum + i
        i = i + 1
    return sum


f()
```


csharp/src/SeedLang/Interpreter/FunctionStack.cs
```diff
@@ -0,0 +1,63 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace SeedLang.Interpreter {
+  // A stack used in the compiler to store compiled functions.
+  internal class FunctionStack {
```
wixette, line 21: This class and its description are a bit hard to understand. And the class name is a bit confusing with `CallStack`. What's a `FunctionStack` required by the compiler? Is the class used for nested functions? If so, I guess it'd be better to name it `NestedFunctionStack`.

csharp/src/SeedLang/Ast/AstStringBuilder.cs
```diff
@@ -211,19 +211,19 @@ internal sealed class AstStringBuilder : AstWalker {
...
 
-    protected override void Visit(FunctionStatement func) {
-      Enter(func);
-      _out.Append($" ({func.Name}:");
+    protected override void Visit(FuncDefStatement funcDef) {
+      Enter(funcDef);
+      _out.Append($" ({funcDef.Name}:");
       bool firstParameter = true;
-      foreach (string parameter in func.Parameters) {
+      foreach (string parameter in funcDef.Parameters) {
```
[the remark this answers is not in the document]
codingpotato: Done

csharp/src/SeedLang/Interpreter/Disassembler.cs
```diff
@@ -12,33 +12,52 @@
...
-        sb.AppendLine();
+      while (_functions.Count > 0) {
+        sb.AppendLine($"Function <{_name}>");
+        for (int i = 0; i < _chunk.Bytecode.Count; ++i) {
+          Debug.Assert(i < _chunk.Ranges.Count);
+          Instruction instr = _chunk.Bytecode[i];
+          Range range = _chunk.Ranges[i];
+          sb.Append($"  {i + 1,-5}{instr.Opcode,-10}{OperandsToString(instr),-15}");
+          string rangeString = range is null ? "" : range.ToString();
+          sb.Append($"  {ConstOperandsToString(instr),-20}{rangeString}");
```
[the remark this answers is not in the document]
codingpotato: Done

csharp/src/SeedLang/Interpreter/Environment.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using SeedLang.Runtime;
+
+namespace SeedLang.Interpreter {
+  // The environment to store names and values of build-in and global variables.
+  //
+  // TODO: handle build-in variables.
+  internal class Environment {
```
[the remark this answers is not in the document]
codingpotato: I have no clear idea about if this `Environment` shall be reused by AST executor to hold the global variables as well, and if I should rename `ScopedEnvironment` to another proper name.
But anyway, I will rename `Environment` to `GlobalEnvironment`. It seems a good choice now.

csharp/src/SeedLang/Interpreter/FunctionStack.cs
codingpotato: Yes, this class is used to hold nested functions, and it's a must, because a main function will be generated for the global code. Then any function definition will be a nested function.
I will rename it `NestedFuncStack`.

csharp/src/SeedLang/Interpreter/FunctionStack.cs
```diff
@@ -0,0 +1,63 @@
+using System;
```
wixette, line 1: The filename of this code is still `FunctionStack.cs`...

codingpotato: Fixed it.

wixette (approved):

PR #121, opened 2022-01-11
Compile and execute if statements in SeedVM

codingpotato:
Following code can be run as byte code:
```python
def fib(n):
  if n <= 2:
    return 1
  else:
    return fib(n - 1) + fib(n - 2)


fib(35)
```


wixette (approved):
LGTM

PR #123, opened 2022-01-15
Compile boolean and comparison expressions

codingpotato:
Following code can be executed by this PR:

```python
if a < b < c and e > f:
   ...
```

### DIFF
--- a/csharp/src/SeedLang/Interpreter/Chunk.cs
+++ b/csharp/src/SeedLang/Interpreter/Chunk.cs
@@ -38,28 +38,28 @@ namespace SeedLang.Interpreter {
     // The constant list to hold all the constants used in this chunk.
     private Value[] _constants;
 
-    internal int Emit(Opcode opcode, uint a, Range range) {
+    // Emits an instruction with the opcode of type A.
+    internal void Emit(Opcode opcode, uint a, Range range) {
       _bytecode.Add(new Instruction(opcode, a));
       _ranges.Add(range);
-      return _bytecode.Count - 1;
     }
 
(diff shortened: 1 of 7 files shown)


csharp/src/SeedLang/Interpreter/Chunk.cs
```diff
@@ -38,24 +38,28 @@ internal class Chunk {
     // The constant list to hold all the constants used in this chunk.
     private Value[] _constants;
 
-    internal void Emit(Opcode opcode, uint a, Range range) {
+    internal int Emit(Opcode opcode, uint a, Range range) {
       _bytecode.Add(new Instruction(opcode, a));
       _ranges.Add(range);
+      return _bytecode.Count - 1;
```
wixette, line 44: If only reading the code of Chunk without looking into the client code, it's really hard to guess why a value of `_bytecode.Count - 1` is always returned.

Given all the Emit functions have the same behavior - returning a value of `_bytecode.Count - 1`, do we really need this return value? Why don't we wrap this logic in the upper/caller function, like:

```
chunk.Emit(...);
int jmpEnd = chunk.Bytecode.Count - 1;
```

Or, maybe to provide another interface for the Chunk class:

```
internal int GetJmpEndPos();
```

Another idea could be using an out parameter to make the purpose of the return value clearer:

```
internal void Emit(..., out int jmpEnd);
```

codingpotato: My intention is to return current code position from Emit methods. But considering the code position is only used by Compiler to calculate jump offsets. So maybe a private function in Compiler to get current code position is enough now.
Fixed as this way. Please have another review again.

wixette (approved):
LGTM

PR #125, opened 2022-02-04
Implement multiple assignment

codingpotato:
Following code is supported by SeedVM:

```python
def fib(n):
  a, b = 0, 1
  i = 1
  while i < n:
    a, b = b, a + b
    i = i + 1
  return b


fib(35)
```


wixette (approved):
LGTM

PR #127, opened 2022-02-08
Implement "for in" statements

codingpotato:
Following code can be executed by SeedVM:

```python
sum = 0
for i in [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]:
  sum = sum + i

sum = 0
for i in range(1, 11):
  sum = sum + i
```

There are more changes in this PR than normal. The most significant changes include:
- Add `for in` statements.
- Execute `for in` statements by AST and SeedVM.
- Refactor Value classes.
- Refactor VariableResolver class.

Please have a review again.


csharp/src/SeedLang/Runtime/NumberRange.cs
```diff
@@ -0,0 +1,48 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace SeedLang.Runtime {
+  // A class to store the range information of "for in range" statements.
+  internal class NumberRange {
```
wixette, line 19: Naming: python simply names this type `range`. Any particular reason you named it `NumberRange`?

csharp/src/SeedLang/Runtime/NativeFunction.cs
```diff
@@ -40,16 +43,77 @@ internal class NativeFunction : IFunction {
 
   // The static class to define all the build-in native functions.
   internal static class NativeFunctions {
+    public const string HasNext = "hasnext";
+    public const string Iter = "iter";
+    public const string Len = "len";
+    public const string List = "list";
```
wixette, line 49: The behavior of `list()` is different with Python:

Python:
```
>>> list(1, 2, 3)
TypeError: list expected at most 1 argument, got 3
```

SeedPython:
```
>>> list(1,2,3)
---------- Source ----------
list(1,2,3)

---------- Run ----------
list(1,2,3)
Eval result: [1, 2, 3]
```

csharp/src/SeedLang/Runtime/NativeFunction.cs
```diff
@@ -40,16 +43,77 @@ internal class NativeFunction : IFunction {
 
   // The static class to define all the build-in native functions.
   internal static class NativeFunctions {
+    public const string HasNext = "hasnext";
+    public const string Iter = "iter";
```
wixette, line 47: Python:
```
>>> iter([1, 2])
<list_iterator object at 0x106cf2730>
```

SeedPython:
```
>>> iter ([1, 2])
---------- Source ----------
iter ([1, 2])

---------- Run ----------
iter ([1, 2])
iter ([1, 2])
: 20220209231748914 Fatal (SeedLang.VM) <> [Ln 1, Col 0 - Ln 1, Col 12] 19: RuntimeErrorInvalidCast
```

csharp/src/SeedLang/Runtime/NativeFunction.cs
```diff
@@ -40,16 +43,77 @@ internal class NativeFunction : IFunction {
 
   // The static class to define all the build-in native functions.
   internal static class NativeFunctions {
+    public const string HasNext = "hasnext";
```
wixette, line 46: What is `hasnext` for? Python doesn't provide this method. Is it for determining the end of an iterator without an exception mechanism?

csharp/src/SeedLang/Runtime/NativeFunction.cs
```diff
@@ -40,16 +43,77 @@ internal class NativeFunction : IFunction {
...
+    public const string Iter = "iter";
+    public const string Len = "len";
+    public const string List = "list";
+    public const string Next = "next";
+    public const string Range = "range";
```
wixette, line 51: Python:
```
>>> list(range(5))
[0, 1, 2, 3, 4]
```

SeedPython:
```
>>> list(range(5))
---------- Source ----------
list(range(5))

---------- Run ----------
list(range(5))
Eval result: [range(0, 5, 1)]
```

It seems SeedPython doesn't evaluate the range when casting it to a list.

csharp/src/SeedLang/Runtime/NumberRange.cs
codingpotato: There are conflicts with Range class in Common component if naming it as Range, we have to write a lot of code like Common.Range. I'm thinking about different options:
 
1. Change the name to RangeValue or something else better.
2. Add a namespace SeedLang.Value and put all value types include Range in it. There are so many value types now. But we still need write Value.Range in the code.
3. Add a namespace SeedLang.Runtime.Value and put all value types include Range in it. There are so many value types now. But we still need write Value.Range in the code.
 
Any ideas about it? Which one do you prefer?

wixette: Ah, good question. I forgot the class name of the source code range.

### Why is `NumberRange` not good?

The name `NumberRange` is leading to an assumption that there are other range classes, such as `StringRange`, `BoolRange`, etc., which do not and will not exist.

### My preference

The 1st idea looks more promising to me. How about a set of Value classes like:

* NumberValue
* StringValue
* BoolValue
* ListValue
* IterValue
* RangeValue
* ...

I am not sure if this naming system is conflicting with your current design.

And in parallel, we might need to have another set of Type classes, like:

* NumberType
* StiringType
* BoolType
* ListType
* IterType
* RangeType
* ...

The could be consistent and easy to understand.

codingpotato: A little problem here, current design is we only have one "struct Value" to hole None, bool, number and HeapObject. The HeapObject will then hole String, Function, List, Iter, Range.

String and List are just C# host objects, Function, Iter and Range are customized heap objects. I think it's not good to define them as XXXObject.

I have a idea that is to define these customized heap objects in the HeapObject class. Let me try if it's a good idea.

csharp/src/SeedLang/Runtime/NativeFunction.cs
codingpotato: Fixed by change the behavior of the list build-in function.

csharp/src/SeedLang/Runtime/NativeFunction.cs
codingpotato: `iter` build-in function is removed, because I realized that there is an easy way to optimize `for in` statements execution, and without the need of `iter`. The new method is several times faster than my previous implementation.

csharp/src/SeedLang/Runtime/NativeFunction.cs
codingpotato: `hasn't` build-in function is removed as well, because of removal of `iter`.

csharp/src/SeedLang/Runtime/NumberRange.cs
codingpotato: Moved Function, Range into HeapObject class.

wixette (approved):
LGTM

PR #128, opened 2022-02-15
Implement attribute statements for build-in value types

codingpotato:
Following code can be run by SeedVM:

```python
array = []
for i in range(5):
  array.append(i)


array
```

### DIFF
--- a/csharp/src/SeedLang/Runtime/NativeFunctions.cs
+++ b/csharp/src/SeedLang/Runtime/NativeFunctions.cs
@@ -27,6 +27,8 @@ namespace SeedLang.Runtime {
     public const string Range = "range";
 
     public static NativeFunction[] Funcs = new NativeFunction[] {
+      // Appends a value to a list. The first argument is the list, the second argument is the
+      // value to be appended to the list.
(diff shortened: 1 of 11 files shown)


csharp/src/SeedLang/Runtime/NativeFunctions.cs
```diff
@@ -21,11 +21,27 @@ namespace SeedLang.Runtime {
...
+        if (length != 2) {
+          throw new DiagnosticException(SystemReporters.SeedRuntime, Severity.Fatal, "", null,
+                                        Message.RuntimeErrorIncorrectArgsCount);
+        }
+        Value value = args[offset];
+        if (value.IsList) {
+          List<Value> list = value.AsList();
+          list.Add(args[offset + 1]);
+        }
+        return new Value(true);
```
wixette, line 42: Here is a difference with Python's built-in function append():

SeedPython
```
>>> a = [].append(1)
---------- Source ----------
a = [].append(1)

---------- Run ----------
>>> a
---------- Source ----------
a

---------- Run ----------
a
Eval result: True
```

Python3
```
>>> a = [].append(1)
>>> a
>>> a is None
True
```

Python's append() function always returns None.

PR #132, opened 2022-02-22
Refactor SeedLang.Shell to show line numbers of source code

codingpotato:
- Show line numbers of source code in SeedLang.Shell, and only show related source lines when displaying evaluation results, please use following command to see the result.

```shell
dotnet run -c Release --project src/SeedLang.Shell -- -f ../examples/seedpython/scripts/sum.py
```

- Small fix for correct range when variables is not defined.
- Small fix for return a None value when execute a non-value return statement.

### DIFF
--- a/csharp/src/SeedLang.Shell/Program.cs
+++ b/csharp/src/SeedLang.Shell/Program.cs
@@ -32,7 +32,7 @@ namespace SeedLang.Shell {
 
       [Option('v', "visualizers", Required = false,
               Default = new VisualizerType[] { VisualizerType.Eval },
-              HelpText = "The Visualizers to be enabled.")]
+              HelpText = "The Visualizers to be enabled. Use All to enable all visualizers.")]
       public IEnumerable<VisualizerType> VisualizerTypes { get; set; }
 
       [Option('f', "file", Required = false, Default = null,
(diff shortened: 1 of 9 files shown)


csharp/src/SeedLang.Shell/Program.cs
```diff
@@ -32,7 +32,7 @@ internal class Options {
 
       [Option('v', "visualizers", Required = false,
               Default = new VisualizerType[] { VisualizerType.Eval },
-              HelpText = "The Visualizers to be enabled.")]
+              HelpText = "The Visualizers to be enabled. Use All to enable all visualizers.")]
```
wixette, line 35: As for this option, it's really hard for users to figure out how to turn on more than one visualizers via commandline.

E.g., it seems the following commandlines work if we'd like to turn on Binary and Eval:

```
dotnet run --project src/SeedLang.Shell -- -v Binary Eval 
dotnet run --project src/SeedLang.Shell -- --visualizers=Binary Eval
```

But the followings do not work:

```
dotnet run --project src/SeedLang.Shell -- --visualizers="Binary Eval"
dotnet run --project src/SeedLang.Shell -- --visualizers="Binary,Eval"
dotnet run --project src/SeedLang.Shell -- --visualizers=Binary,Eval 
```

Can we have a better HelpText to describe the correct usage of providing more than one visualizers?

csharp/src/SeedLang/Runtime/RunType.cs
```diff
@@ -15,9 +15,13 @@
...
   public enum RunType {
-    // Parses the source code into an AST tree, and runs it by traversing the AST tree.
+    // Parses source code into an AST tree, and runs it by traversing the AST tree.
     Ast,
-    // Parses and compiles the source code into bytecode, and runs it in a VM.
+    // Parses and compiles source code into bytecode, and runs it in a VM.
     Bytecode,
+    // Dumps source code as AST tree.
+    DumpAst,
+    // Disassembles source code to bytecode.
```
wixette, line 24:
```suggestion
    // Compiles source code to bytecode then prints the disassembled bytecode.
```

codingpotato: Done

csharp/src/SeedLang.Shell/Program.cs
codingpotato: Add help text description manually, because the CommandLineParser couldn't generate help text for sequence type options.
Also change the separator to "," which might be better than the space.

wixette (approved):
LGTM

PR #133, opened 2022-02-22
Sorting examples in SeedPython

wixette:
This PR does not need to be merged now, since the code files are not 100% supported by the current state of SeedPython.

I made this PR to show the expected state that we'd like to support a sorting algorithm visualization app. We can use this PR to discuss what we'd like to support and what not to support in SeedLang for building such an app.


wixette:
@codingpotato I updated this PR to take the current design of V-tags. Although there are still some features that the current SeedPyhton does not support (See the issues I reported: #166, #168, #169), shall we probably merge this PR into the main branch, so that we can improve V-tags based on the sorting algorithms listed in this PR?

codingpotato (approved):
LGTM

PR #138, opened 2022-02-27
Implement the print build-in function

codingpotato:
- Implement `print`
- Remove `EVAL` opcode, and add an `__printval__` build-in function, which is used for SeedCalc and SeedX to support interactive mode evaluation. The behavior is almost the same as Python now.
- Add Stdout in Sys class to support `print` and `__printval__` build-in functions.


csharp/src/SeedLang/Runtime/Events.cs
```diff
@@ -113,4 +113,13 @@ public class EvalEvent : AbstractEvent {
       Value = value;
     }
   }
+
+  // An event which is triggered when an print build-in function is executed.
+  public class PrintEvent : AbstractEvent {
```
wixette, line 118: Do we really need a separate visualizer event for every single built-in function?

From what we see in this code, we already have `EvalEvent` for `eval`, `PrintEvent` for `print`... Will every built-in function require an event finally?

Is it possible to have an event such as `BuiltInFuncCallEvent`, and have the function name and function arguments as its parameters?

csharp/src/SeedLang/Runtime/HeapObjects.cs
```diff
@@ -15,31 +15,30 @@
 using System;
 
 namespace SeedLang.Runtime {
-  using NativeFunctionType = Func<Value[], int, int, Value>;
+  using BuildinFunctionType = Func<Value[], int, int, VisualizerCenter, Common.Range, Value>;
 
   internal partial class HeapObject {
-    // The function interface of function value types.
+    // A empty interface for all function value types. It's only used to identify function types.
```
wixette, line 21:
```suggestion
    // An empty interface for all function value types. It's only used to identify function types.
```

csharp/src/SeedLang/Runtime/NativeFunctions.cs
```diff
@@ -17,73 +17,115 @@
...
+      return new Value();
+    }
 
-      // Creates an empty list if the length of arguments is empty, and a list if the argument is a
-      // subscriptable value.
-      new NativeFunction(List, (Value[] args, int offset, int length) => {
-        if (length < 0 || length > 1) {
+    // Evaluates a value when it's not none. It's used by SeedCalc and interactive mode SeedX
+    // primarily.
+    private static Value EvalFunc(Value[] args, int offset, int length, VisualizerCenter vc,
```
wixette, line 57: Python3's `eval` function accepts a string as its parameter. This is different from our current implementation, which actually accepts a runtime value/expression as its parameter.

```
python3
Python 3.9.10 (main, Jan 15 2022, 11:48:04) 
[Clang 13.0.0 (clang-1300.0.29.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> eval(3+4)
TypeError: eval() arg 1 must be a string, bytes or code object
>>> eval("3+4")
7
```

```
dotnet run --project src/SeedLang.Shell
SeedLang.Shell 0.2.0-preview20220227142557

>>> eval(3+4)
---------- Source ----------
1     eval(3+4)
---------- Run ----------
1     eval(3+4)
Eval: 7
```


Do we plan to make our `eval` function consistent with Python's design?

If the answer is yes, we should at least add a TODO comment here.

If it's not the case, what's the reason we need to expose this function to users? What can our users do with this function? I think in SeedLang.Shell we can implicitly call this function without letting our users know that, can't we?

And, if we really need an `eval` function that accepts a runtime value (expression) as its parameter,  can we name this built-in function slightly different from Python's? Probably we can name it `valeval` which means "value evaluation".

csharp/src/SeedLang/Runtime/NativeFunctions.cs
```diff
@@ -17,73 +17,115 @@
...
+        return args[offset];
+      }
+      var list = new List<Value>();
+      for (int i = 0; i < args[offset].Length; i++) {
+        list.Add(args[offset][i]);
+      }
+      return new Value(list);
+    }
+
+    private static Value PrintFunc(Value[] args, int offset, int length, VisualizerCenter vc,
```
wixette, line 101: I think `print()` is not a function that does not do anything but notifies registered visualizers about the parameters.

A typical semantic meaning of `print` should be print contents to the standard output or the debug console.

Are we going to implement the real feature of `print`? Are we going to need something like Python's `__str__` as well to support `print`?

codingpotato: The reason that I implement `print` as a notification to the client is SeedLang core doesn't know how to print. It might be calling `Console.Write` in SeedLang.Shell, or output to a Unity 3D log widget in the Sorting App. Then notification can be used to separate SeedLang core and the client.
Another approach could be just writing to the stdout, and let the client capture the content from the stdout. It might be a little difficult to handle.
@wixette Any thoughts about these two options?

__str__ can be implement later as an attribute of string value type. Current Value.AsString() method can be used to support `print`.

csharp/src/SeedLang/Runtime/NativeFunctions.cs
codingpotato: This function is not intended to be `eval`. It used to handle the single expression statement, and output the value. I think only SeedCalc and SeedLan.Shell need it.
We need this function because compiler need place It in the byte code to print the value of expression statement. SeedLang.Shell doesn't know when to call it.
How about change it to a internal build-in function like `__evalval__`?

For the real `eval` function, we can decide if we need it later.

csharp/src/SeedLang/Runtime/NativeFunctions.cs
wixette: print() is so common in script languages thus it is worth supporting it in different environments. A good and similar example is JavaScript's `console.log()`: In node.js commandline env, `console.log()` outputs to stdout since the console in the context is bound to stdout; while in a browser's window, `console.log()` outputs to browser's developer tools window, since the console object in this context is bound to the browser window.

I guess we need to support this binding mechanism for Python's `print` and  Lua's `print` too.

1. We will probably need a built-in library object `sys.stdout` (in SeedPython) or `io.stdout` (in SeedLua), that is bound to stdout by default. SeedLang's users can have an interface provided by SeedLang runtime to re-bind `sys.stdout` or `io.stdout` to another object. In this way, the client can easily customize the behavior of the `print` function.

2. If we do not plan to introduce `sys.stdout` and `io.stdout` for now, we can make the object hidden from the client code. E.g., to have an internal built-in object named `__stdout__`. Then, we provide an interface called `RedirectStdout(...)` in SeedLang's public interface.

3. Since we are building SeedLang in C#, the interface of `RedirectStdout(...)` could be something like below:

```
namespace SeedLang.Runtime {
  class ... {
    void RedirectStdout(System.IO.TextWriter writer, out System.IO.TextWriter oldWriter);
  }
}
```

In this way, users can implement their own TextWriter class to accept the text to be printed.

csharp/src/SeedLang/Runtime/NativeFunctions.cs
wixette: `__evalval__` sounds perfect to me.

csharp/src/SeedLang/Runtime/Events.cs
wixette: Two different but related things are being mixed up here.

1. The intrinsic or expected behavior of functions that make side effects to the system (via inputting or outputting): Functions like `print` or `io.stderr.write` or `console.log` are designed to make side effects to the system.

2. The visualizer framework to notify the outer world about what is happening in the SeedLang runtime: e.g., a function call event can be passed out to visualizers to notify the thing that is happening.

I think both two things are valuable to users and it might not be a good choice to mix them up.

csharp/src/SeedLang/Runtime/NativeFunctions.cs
codingpotato: Rename it `__printval__` eventually, because this function doesn't evaluate anything, it just print the string representative of a value if it's not none in the interactive mode.

wixette (approved):
LGTM

PR #141, opened 2022-03-06
Suport string operations in SeedVM

### LINKED ISSUE #126: Inconsistencies between SeedPython and Python
Let's use this issue to track and discuss the inconsistent cases between the two languages: SeedPython and Python.

codingpotato:
- Add string support in SeedVM. All code in `string.py` in the example folder can be run.
- Implement more number operators. All code in `number.py` in the example folder can be run.
- Fix #126


wixette (approved):
LGTM

PR #148, opened 2022-03-19
Implement Dict

codingpotato:
- Implement Dict for SeedPython, support following code now

```python
dict = {'Name': 'Zara', 'Age': 7, 'Class': 'First'}

print("dict['Name']:", dict['Name'])  # dict['Name']: Zara
print("dict['Age']:", dict['Age'])  # "dict['Age']: 7

dict['Age'] = 8
dict['School'] = "DPS School"

print("dict['Age']:", dict['Age'])  # dict['Age']: 8
print("dict['School']:", dict['School'])  # dict['School']: DPS School
```

- Change the underlying data structure of Tuple to support hash code calculation for Dict.

### DIFF
--- a/csharp/src/SeedLang/Ast/Expressions.cs
+++ b/csharp/src/SeedLang/Ast/Expressions.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 // Copyright 2021-2022 The SeedV Lab.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -68,9 +69,8 @@ namespace SeedLang.Ast {
     }
 
     // The factory method to create a dictionary expression.
-    internal static DictExpression Dict(Expression[] keys, Expression[] values, Range range) {
-      Debug.Assert(keys.Length == values.Length);
-      return new DictExpression(keys, values, range);
+    internal static DictExpression Dict(KeyValuePair<Expression, Expression>[] items, Range range) {
+      return new DictExpression(items, range);
     }
 
     // The factory method to create a list expression.
@@ -188,12 +188,11 @@ namespace SeedLang.Ast {
   }
 
   internal class DictExpression : Expression {
-    public Expression[] Keys { get; }
-    public Expression[] Values { get; }
+    public KeyValuePair<Expression, Expression>[] Items { get; }
 
-    internal DictExpression(Expression[] keys, Expression[] values, Range range) : base(range) {
-      Keys = keys;
-      Values = values;
+    internal DictExpression(KeyValuePair<Expression, Expression>[] items, Range range) :
+        base(range) {
+      Items = items;
     }
   }
 
--- a/csharp/src/SeedLang/Ast/Expressions.cs
+++ b/csharp/src/SeedLang/Ast/Expressions.cs
@@ -67,6 +67,12 @@ namespace SeedLang.Ast {
       return new StringConstantExpression(ValueHelper.Unescape(value), range);
     }
 
+    // The factory method to create a dictionary expression.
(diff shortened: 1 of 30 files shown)


csharp/src/SeedLang/Runtime/HeapObject.cs
```diff
@@ -222,12 +270,20 @@ internal partial class HeapObject : IEquatable<HeapObject> {
       return intIndex < 0 ? length + intIndex : intIndex;
     }
 
-    private static string TupleToString(IReadOnlyList<Value> tuple) {
+    private static void CheckKey(Value key) {
+      if (!key.IsNil && !key.IsBoolean && !key.IsNumber && !key.IsString && !key.IsRange &&
+          !key.IsTuple) {
+        throw new DiagnosticException(SystemReporters.SeedRuntime, Severity.Fatal, "", null,
+                                      Message.RuntimeErrorUnhashableType);
```
wixette, line 277: It seems RuntimeErrorUnhashableTypes should also be checked when the dictionary value is initiated. For example, the current SeedPython implementation outputs as below:

```
>>> a = { [1, 2, 3]: 1 }
---------- Source ----------
1     a = { [1, 2, 3]: 1 }
---------- Run ----------
>>> print(a)
---------- Source ----------
1     print(a)
---------- Run ----------
{[1, 2, 3]: 1}
>>> a
---------- Source ----------
1     a
---------- Run ----------
{[1, 2, 3]: 1}
>>> a[[1, 2, 3]]
---------- Source ----------
1     a[[1, 2, 3]]
---------- Run ----------
1     a[[1, 2, 3]]
20220319211523331 Fatal (SeedLang.VM) <> [Ln 1, Col 0 - Ln 1, Col 11] 29: RuntimeErrorUnhashableType
```

While Python3 reports Unhashable key error as early as the value is created:

```
Python 3.9.10 (main, Jan 15 2022, 11:48:04) 
[Clang 13.0.0 (clang-1300.0.29.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> a = { [1, 2, 3]: 3 }
TypeError: unhashable type: 'list'
```

csharp/src/SeedLang/Runtime/HeapObject.cs
```diff
@@ -222,12 +270,20 @@ internal partial class HeapObject : IEquatable<HeapObject> {
...
+      }
+    }
+
+    private static string TupleToString(ImmutableArray<Value> tuple) {
       var sb = new StringBuilder();
       sb.Append('(');
-      for (int i = 0; i < tuple.Count; i++) {
+      for (int i = 0; i < tuple.Length; i++) {
```
wixette, line 284: I guess ImutableArray can be passed into String.Join so that the following lines can be replaced with a single line:

```
sb.Append(String.Join(", ", tuple));
```

csharp/src/SeedLang/Ast/Expressions.cs
```diff
@@ -181,6 +187,16 @@ internal class StringConstantExpression : Expression {
     }
   }
 
+  internal class DictExpression : Expression {
```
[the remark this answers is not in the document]
codingpotato: Use CSharp build-in `KeyValuePair` structure to improve it.

csharp/src/SeedLang/Runtime/HeapObject.cs
codingpotato: Fix it by adding check in the constructor.

csharp/src/SeedLang/Runtime/HeapObject.cs
codingpotato: Fixed.

wixette (approved):
LGTM

PR #149, opened 2022-03-21
Implement in keyword

codingpotato:
Support following code:

```python
print('python' in str)
print('super' in str)

print(8 in list)
print(None in list)

print('Age' in dict)
print('School' in dict)
```


wixette (approved):
LGTM

PR #152, opened 2022-03-26
Implement comparison notification

codingpotato:
Run following command to see the result of this PR:

```shell
dotnet run --project csharp/src/SeedLang.Shell -- -v All -f examples/seedpython/scripts/bubble_sort.py

```

### DIFF
--- a/csharp/src/SeedLang/Interpreter/Compiler.cs
+++ b/csharp/src/SeedLang/Interpreter/Compiler.cs
@@ -22,7 +22,7 @@ namespace SeedLang.Interpreter {
   // The compiler to convert an AST tree to bytecode.
   internal class Compiler : AstWalker {
     // The class to collect comparison information during compilation. The information includes
-    // registers to hold the operand values and operators, which is used to generate comparison
+    // operators and registers to hold the operand values, which is used to generate comparison
     // notifications.
     private class ComparisonInfo {
       public uint FirstId;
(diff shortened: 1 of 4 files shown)


csharp/src/SeedLang/Interpreter/Compiler.cs
```diff
@@ -21,11 +21,21 @@
 namespace SeedLang.Interpreter {
   // The compiler to convert an AST tree to bytecode.
   internal class Compiler : AstWalker {
+    // The class to collect comparison information during compilation. The information includes
+    // operators and registers to hold the operand values, which is used to generate comparison
+    // notifications.
+    private class ComparisonInfo {
+      public uint FirstId;
```
wixette, line 28: 1. The following value IDs are named `ValueIds`. Better to rename this to `FirstValueId` to make them consistent.

3. Although I figured out what `firstId` is for, I think it'd be better to add some very simple comments to let other developers know that it's for holding the first value ID of a comparison chain, such as `3 > 2 > 1 > 4`

csharp/src/SeedLang/Interpreter/Compiler.cs
```diff
@@ -21,11 +21,21 @@
...
+      public List<ComparisonOperator> Ops = new List<ComparisonOperator>();
+      public List<uint> ValueIds = new List<uint>();
+    }
```
wixette, line 31: As we discussed before, the most tricky problem for a sorting algorithm visualizer about the comparison event is:

For a sorting algorithm visualizer, it's way useful to know what variables are being compared, instead of what evaluated values are being compared.

For example, as for the comparison `if a[i] > a[j]:` which is very common in compare-and-swap sorting algorithms, the sorting visualizer doesn't care much about the actual value of `a[i]` and `a[j]`. Instead, the visualizer cares very much about the semantic meaning of `a[i]` and `a[j]` themself - the `i`th element and the `j`th element of the data list `a`.

If it's not easy to pass semantic info out to the visualizer via notification event, it's still valid to use `V-Tag` to express additional info for a comparison event, such as:

```
# [[ Compare(a, i, j) ]]
if a[i] > a[j]:
  ...
```

codingpotato: @wixette there are multiple cases for comparison expressions, for example:

```python
if 1 > 2:
if a > b:
if a + 1 > 100 / 3:
if a[1] < c:
if a[i] * 5 != b[j] * 6:
if a == True:
if a[i] < a[j]:
if i in (1, 2, 3):
```

It's hard to know which semantic info need for the visualizers. Maybe V-Tag can be used here.
And we need decide do we need support this kind of comparison notification or we only need support V-Tag for comparison expressions.

PR #160, opened 2022-04-04
Implement single line VTag parsing and executing

codingpotato:
Support single line VTag:

```python
array = [64, 34, 25, 12, 22, 11, 90]
n = len(array)
for i in range(n):
    for j in range(n - i - 1):
        if array[j] > array[j + 1]:
            # [[ Swap ]]
            array[j], array[j + 1] = array[j + 1], array[j]


print(array)
```


csharp/src/SeedLang/X/SeedPythonDentLexer.cs
```diff
@@ -145,27 +148,32 @@ internal class SeedPythonDentLexer : SeedPythonLexer {
...
-        return false;
+    private void ScanVTag(IToken comment) {
+      if (comment.Text.Length > 1) {
+        Debug.Assert(comment.Text[0] == '#');
+        var inputStream = new AntlrInputStream(comment.Text.Substring(1));
+        var lexer = new SeedPythonLexer(inputStream);
+        var tokens = lexer.GetAllTokens();
+        if (tokens.Count > 0 && tokens[0].Type == SeedPythonParser.VTAG_START) {
+          for (int i = 0; i < tokens.Count; i++) {
+            // Adds back '#' for the VTAG_START token.
```
wixette, line 159: Please double-check if the following adding back logic is meeting the expectation:

Here is a common case:

```
    public void TestVTag() {
      string source = "# \t\t\t\t[[ Print ]]\nprint(1)";
      var expectedTokens = new string[] {
          $"[@-1,0:7='# [[',<{SeedPythonParser.VTAG_START}>,1:0]",
          ... ...
      };
```

I added 4 "\t" to the above test case between "# " and "[[". The first output token has its range set to 0:7 but its actual length is 4.

codingpotato: Great catch. Fixed this bug.

wixette (approved):
LGTM

PR #165, opened 2022-04-09
Implement multiple-line VTag statements and arguments

codingpotato:
Use following command to run bubble_sort.py to check the result of Tag statements

```shell
dotnet run --project csharp/src/SeedLang.Shell -- -v All -f examples/seedpython/scripts/bubble_sort.py
```


wixette (approved):
LGTM

PR #172, opened 2022-04-17
Implement FuncCalled and FuncReturned visualization events

codingpotato:
Also refactor VisualizerManager by reflection.

### DIFF
--- a/csharp/src/SeedLang.Shell/Executor.cs
+++ b/csharp/src/SeedLang.Shell/Executor.cs
@@ -13,19 +13,17 @@
 // limitations under the License.
 
 using System;
-using System.Collections.Generic;
 using System.IO;
 using SeedLang.Common;
 
 namespace SeedLang.Shell {
   internal static class Executor {
-    internal static void Repl(SeedXLanguage language, RunType runType,
-                              IEnumerable<VisualizerType> visualizerTypes) {
+    internal static void Repl(SeedXLanguage language, RunType runType) {
       ReadLine.HistoryEnabled = true;
       var engine = new Engine(language, RunMode.Interactive);
-      var source = new SourceCode();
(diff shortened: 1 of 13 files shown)


wixette (commented):
```
dotnet run --project src/SeedLang.Shell -- -v "Func*"
SeedLang.Shell 0.2.0-preview20220417091507

Enabled Visualizers: FuncCalled, FuncReturned

>>> def foo(m):
...   print(m)
... 
---------- Source ----------
1     def foo(m):
2       print(m)
3     

---------- Run ----------
>>> 
>>> 
>>> 
>>> foo(3)
---------- Source ----------
1     foo(3)

---------- Run ----------
1     foo(3)
FuncCalled: foo 3
FuncCalled: print 3
3
FuncCalled: __printval__ None
FuncReturned: foo None
1     foo(3)
FuncCalled: __printval__ None
>>>
```

Three quick questions:

1. Can we turn off sending out Func* events for internal functions such as `__printval__`?

2. For the`foo()` function in the above code, there were a `FuncCalled` event and a `FuncReturned` event, which is great. But why were there two `FuncCalled` events for `__printval__`? It should have been called only once, right?

3.  It's good to also have `FuncCalled` event for built-in functions like `print()`. But why doesn't it have a paired`FuncReturned` event? What's the expected behavior for built-in functions like `print()`?

codingpotato:
1. Implement it. Will not send events for internal functions.
2. In this case, `__printval__` will be called twice. First one is to print the return value of `print()` function. Second one is to print the return value of `foo()` function.
3. Optimize the design to implement it.

wixette (approved):
I tested the FuncCalled and the FuncReturned events for a fib() function. The output is super cool!

LGTM.

PR #173, opened 2022-04-18
Extract SeedBlock-related code to experimental dir

wixette:
Since we have decided to avoid SeedLang's dependency on SeedBlock since the design of SeedBlock will be refined a lot, I am trying to extract all the SeedBlock-related code and isolate the code under the experimental/ dir. For making the rest of the SeedLang code work, I also made a couple of simple code changes.

Is this a good solution for the existing SeedBlock code? If so, I can update the existing design docs accordingly later.


codingpotato (approved):
LGTM

PR #175, opened 2022-04-20
Remove the Range abstract class

codingpotato:
Also change the TextPosition from struct to sealed class, because the objects of TextPosition could be shared without copy. It might increase the performance a little.

Also changed GetHashCode() methods of TextPosition and TextRange based on the discussion of: https://stackoverflow.com/questions/263400/what-is-the-best-algorithm-for-overriding-gethashcode


wixette (approved):
LGTM

PR #178, opened 2022-04-24
Refactor the Parser and Compiler classes

codingpotato:
Refactor some code to separate `Compiler` and `ExprCompiler`, which can make future development easier.

There isn't any functional change in the PR.


csharp/src/SeedLang/Ast/AstDumper.cs
```diff
@@ -0,0 +1,319 @@
...
+          return "%";
+        default:
+          throw new NotImplementedException($"Unsupported binary operator: {op}.");
+      }
+    }
+
+    // Returns the internal string representation of boolean operators.
+    internal static string Symbol(this BooleanOperator op) {
+      switch (op) {
+        case BooleanOperator.And:
```
wixette, line 46: "And", "Or" and "Not: both Python and Lua use lowercase form for those operators - "and", "or", "not". Shall we follow this convention for internal representations of them?

csharp/src/SeedLang/Engine.cs
```diff
@@ -34,8 +34,8 @@ public class Engine {
     private readonly RunMode _runMode;
     private readonly VM _vm = new VM();
 
-    // The AST tree of the source code.
-    private AstNode _astNode;
+    // The AST tree of the program.
+    private Statement _program;
```
wixette, line 38: Why don't name this variable "_astTree" or "_programAsAst" ?

codingpotato: Fixed.

csharp/src/SeedLang/Ast/AstDumper.cs
codingpotato: Fixed.

csharp/src/SeedLang/Ast/AstDumper.cs
```diff
@@ -0,0 +1,319 @@
...
+
+    // Returns the internal string representation of unary operators.
+    internal static string Symbol(this UnaryOperator op) {
+      switch (op) {
+        case UnaryOperator.Positive:
+          return "+";
+        case UnaryOperator.Negative:
+          return "-";
+        case UnaryOperator.Not:
+          return "Not";
```
wixette, line 85:
```suggestion
          return "not";
```

wixette (approved):
LGTM

PR #179, opened 2022-04-29
Implement single step notifications

codingpotato:
Please run following command to see the example of this feature:

```shell
dotnet run --project csharp/src/SeedLang.Shell -- -v SingleStep -f examples/seedpython/scripts/sorting/quick_sort.py
```


csharp/src/SeedLang/Interpreter/CompilerHelper.cs
```diff
@@ -218,5 +261,36 @@ internal class CompilerHelper {
...
+
+    private static bool CanFollowSingleStep(Opcode opcode) {
+      switch (opcode) {
+        case Opcode.LOADBOOL:
+        case Opcode.EQ:
+        case Opcode.LT:
+        case Opcode.LE:
+        case Opcode.IN:
+        case Opcode.TEST:
+        case Opcode.TESTSET:
```
wixette, line 273: It's a bit challenging for readers to understand why and how these opcodes are excluded from single step notification emitting. For example,

1. If some source code is compiled to the following instruction sequence:

```
TEST if R(A) == C then PC++     ;; Source code line L1
JMP sBx                         ;; Source code line L1 - same as above
... ...                         ;; Source code line L2
```

Since the `TEST` instruction is excluded, will the single step notification be sent out at the `JMP` instruction for the source code line L1?

What's the root reason that all branch instructions are excluded from single step notification emitting?

I think it'd be way better to add some tech comments for the method CanFollowSingleStep to explain the underlying considerations.

2. I see `VISNOTIFY` instruction is also emitted by the same method group which include the TryEmitSingleStepNotification() call. Shouldn't the `VISNOTIFY` instruction be excluded from single step notification as well?

codingpotato: 1. There is no problem for the normal jump instruction, the problem only happens for conditional `PC++` instruction, e.g.
```text
1. TEST if R(A) == C then PC++
2. JMP sBx
3. ...
```
The execution path is:
```text
1. if R(A) == C then PC++, and VM also performs PC++ as well, so jump to 3
2. otherwise VM performs PC++, will execute 2
```
If we insert a VISNOTIFY between instruction 1 and 2:
```text
1. TEST if R(A) == C then PC++
2. VISNOTIFY
2. JMP sBx
3. ...
```
The original execution path will be destroy in this situation.

But after re-consider clearly, I found instruction 1 and 2 will always belong to the same line even there is a `\` to divide them into two lines, so that it's not possible that a VISNOTIFY instruction is inserted.
So I deleted the logic to ignore these opcodes.

2. Great catch. Fixed this bug in the new commit.

wixette (approved):
LGTM

PR #184, opened 2022-05-03
Add more interfaces in the notified value class

### LINKED ISSUE #183: How to access list elements in the handler of Event.Assignment and Event.VTagExited?
Both `Event.Assignment` and `Event.VTagExited` use `IValue` to pass out variables' values. E.g.:

```
  public class Assignment : AbstractEvent {
      public string Name { get; }
      public VariableType Type { get; }
      public IValue Value { get; }
  }
```

Given the current definition of IValue:

```
  public interface IValue {
    bool IsNil { get; }
    bool IsBoolean { get; }
    bool IsNumber { get; }
    bool IsString { get; }
    bool IsList { get; }
    bool IsFunction { get; }

    bool Boolean { get; }
    double Number { get; }
    string String { get; }
    int Length { get; }
  }
```

How does the client code get the element values of a list value? There is neither `ToList` conversion nor `GetAt` method when `IValue` is a list...

This is pretty similar to the requirement of a typical debugger's inspector: during a breakpoint of a program, the inspector needs to read the element values of a list value that is hold by a variable in the current execution scope.

codingpotato:
- Add a new `SeeldLang.Visualization` namespace.
- Rename the `Value` struct to `VMValue`.
- Create a new `Value class` in `SeeldLang.Visualization` namespace to wrap VMValue, and add interfaces to access `List` and `Dict` elements.

Fix #183


csharp/src/SeedLang/Visualization/Value.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    public bool IsDict => _value.IsDict;
+    public bool IsList => _value.IsList;
+    public bool IsTuple => _value.IsTuple;
+
+    public int Length => _value.Length;
+
+    private readonly VMValue _value;
+
+    public Value() {
```
wixette, line 36: Do we really need these public constructors for visualizers to instantiate a Value? What are the relevant use cases?

```
    public Value() {
      _value = new VMValue();
    }

    public Value(bool boolean) {
      _value = new VMValue(boolean);
    }

    public Value(double number) {
      _value = new VMValue(number);
    }

    public Value(string str) {
      _value = new VMValue(str);
    }
```

A potential use case could be that a visualizer has the ability to update the value of a variable during runtime. It is a complex case. I think we can probably postpone this "live-update" feature to some time later.

I think in most common cases, visualizers simply read the value of a runtime value, and do not need constructors of the Value class.

csharp/src/SeedLang/Visualization/Value.cs
```diff
@@ -0,0 +1,91 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using SeedLang.Runtime;
+
+namespace SeedLang.Visualization {
+  // A wrapper class of the underlying VMValue. It's used to represent runtime values in the
```
wixette, line 19: I like this design to separate the internal representation (VMValue) and the external representation (Value).

csharp/src/SeedLang/Visualization/Value.cs
```diff
@@ -0,0 +1,91 @@
...
+      _value = new VMValue(boolean);
+    }
+
+    public Value(double number) {
+      _value = new VMValue(number);
+    }
+
+    public Value(string str) {
+      _value = new VMValue(str);
+    }
```
wixette, line 50: In case we need a few constructors for the client code to instantiate a value themselves, do we need the constructors to instantiate a value from a list or a dictionary as well? E.g.:

```
    public Value(IReadOnlyList<Value> ...) {
      _value = ...
    }
```

csharp/src/SeedLang/Visualization/Value.cs
codingpotato: One scenario is used as a key or index for the subscript operation. For example, if the user get a VTag notification like `Swap(array, i, j)` or `Access(dict, key)`, he can use `array[i]` or `dict[key]`. But he has to create a value if he world like to access other elements, like:
```python
array[new Value(1)]
dict[new Value("name")]
```
@wixette Let's decide if this use case need be support now.

csharp/src/SeedLang/Visualization/Value.cs
codingpotato: The constructors of Tuple, List and Dict might be needed in the future. But I don't know if we really need it now. Let's postpone them?

csharp/src/SeedLang/Visualization/Value.cs
wixette: Ah, you're right. We need those constructors to support index-access and key-access.

csharp/src/SeedLang/Visualization/Value.cs
wixette: True. We will need constructors for Tuple/List/Dict sometime later but we can postpone it somehow.

wixette (approved):
LGTM

Issue #151, opened 2022-03-25
Notify chained name for variables


codingpotato:
@wixette Just want to confirm that we really need chained name for variables, for example, `Global.var1`, `Global.func1.var1`. They shall be static scoping name chains. We need implement some mechanism to collect the chained name when variables are defined.

How about class and instance variable after we implement Class? Shall we use `Global.class1.var1` and `Global.func1.object1.var1` as the chained name? Is there any confusion here?

codingpotato:
The static scope doesn't match the dynamic scope. For example, in the recursive call case:
```python
def f(n):
  if n == 1:
    return 1
  else:
    return n * f(n - 1)


f(10)
```
During execution, there will be 10 `n` in the stack. The static scope of them are all f.n. The client still cannot tell which `n` it is? At any time only one local `n` can be access by the client.
@wixette I'm just thinking about if the static scope name is really needed by the client?

wixette:
My original request contains the case of embedded static scopes. For example, in Python:

```
m = 1
class F:
   m = 2
   def foo(self):
     m = 3
     print(m)
```

There are 3 `m` variables in the above code, located at different embedded scopes. We need some mechanism to figure out which `m` is being accessed in visualizers.

If counting Python's embedded functions in, there will be more levels of embedded scopes.

And, also consider my above example:

```
def foo(x, y):
  a = x + y
  return a

def bar(x, y):
  a = x * y
  return a
```

How can we mark variable `a` in different function scopes?

Dynamic scope is not needed to be marked here I thinks. Only static scope is good enough.

codingpotato:
Closed by #186

PR #186, opened 2022-05-04
Implement subscript assignment notification

### LINKED ISSUE #182: No assignment event sent out for lists' index assignment operations
```
dotnet run --project src/SeedLang.Shell -- -v Assignment
SeedLang.Shell 0.2.1-preview20220502124751

Enabled Visualizers: Assignment

>>> a = [1, 2, 3]
---------- Source ----------
1     a = [1, 2, 3]

---------- Run ----------
1     a = [1, 2, 3]
Assign global a = [1, 2, 3]
>>> a[0] = 3
---------- Source ----------
1     a[0] = 3

---------- Run ----------
```

There is no notification for the code line `a[0] = 3`.

codingpotato:
Fix #182

### DIFF
--- a/csharp/src/SeedLang/Visualization/Events.cs
+++ b/csharp/src/SeedLang/Visualization/Events.cs
@@ -18,6 +18,7 @@ using SeedLang.Common;
 using SeedLang.Runtime;
 
 namespace SeedLang.Visualization {
+  // The type of variables.
   public enum VariableType {
     Global,
     Local,
@@ -33,9 +34,11 @@ namespace SeedLang.Visualization {
   }
 
   public static class Event {
-    // An event which is triggered when an assignment statement is executed.
+    // An event which is triggered when a value is assigned to a variable.
     public class Assignment : AbstractEvent {
+      // The name of the assigned variable.
       public string Name { get; }
+      // The type of the assigned variable.
       public VariableType Type { get; }
       public Value Value { get; }
 
@@ -72,8 +75,8 @@ namespace SeedLang.Visualization {
       public IReadOnlyList<Value> Values { get; }
       public Value Result { get; }
 
-      public Boolean(BooleanOperator op, IReadOnlyList<Value> values, Value result,
-                     TextRange range) : base(range) {
+      public Boolean(BooleanOperator op, IReadOnlyList<Value> values, Value result, TextRange range)
+          : base(range) {
         Debug.Assert(values.Count > 1);
         Op = op;
         Values = values;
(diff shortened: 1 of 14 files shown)


csharp/src/SeedLang/Visualization/Events.cs
```diff
@@ -128,6 +131,31 @@ public class SingleStep : AbstractEvent {
       public SingleStep(TextRange range) : base(range) { }
     }
 
+    // An event which is triggered when a value is assigned to an element of a container.
+    //
+    // The container might be unnamed variables in following cases. The name of the container is
+    // null and the type is undefined in these cases.
+    // 1) Sets the element of a temporary container: [1, 2, 3][1] = 5
+    // 2) Sets the element of a intermediate container: a[1][2] = 5
```
wixette, line 139: This is very tricky for the 2-dimensional matrix access, such as:

```
matrix2x3 = [ [1, 2], [3, 4], [5, 6]]
matrix2x3[1][1] = 9
```

This is a common case in basic algorithms or data structure programs.

The current implementation only outputs the following event:

```
SubscriptAssign: [3, 9][1] = 9
```

What's the meaning of `3, 9` here?

It's hard for the visualizer to tell which variable is accessed.

Is it possible to divide the operation `matrix2x3[1][1] = 9` into two phases:

- phase 1: get the left value by accessing the memory position of `matrix2x3[1]`, and assign the position to an `unnamed` variable.
- phase 2: assign `9` to the `unnamed` variable.

And send out the info of both the two phases to the client side?

As for a way more complex case, like:

```
matrix[3][4][a][b][c][9][8] = 10
```

It's kind of a chained memory access operation. Can we send out the full memory access chain out so that the client understands where the value is actually put in?

This is really a complex problem. We might need more considerations and experiments here.

Also consider how a typical modern debugger/insepctor handles the code line:

```
matrix[3][4][a][b][c][9][8] = 10
```

Does a modern debugger/inspector know where in the matrix list an internal element is being changed?

codingpotato: New changes for this PR:
1. Refactor VariableResolver class to support chained names of variables. Fixed #151 
2. Remove block scope temporarily to make variable scoping is similar with Python.
3. Try to collect all information of multi-dimensional matrix, but found these information cannot be collected during compile time. It can only be collected during runtime. So only 1-dimensional array elements assignment notification is supported in this PR (Temporary array is also ignored). I will add an issue to consider how to track the reference of all array elements based on registers information during runtime.
4. From this conversation, I think maybe the most interested things of SeedLang clients are changes related named variables and elements of containers. So we can only focus on these changes, e.g. comparison of named variables and container elements, send notification of them, and ignore comparison of const of temporary variables. Is it reasonable?
5. Add `FluentAssertions` into unit test project to make testing easier.

Please have a review again.

wixette: > Try to collect all information of multi-dimensional matrix, but found these information cannot be collected during compile time. It can only be collected during runtime. So only 1-dimensional array elements assignment notification is supported in this PR (Temporary array is also ignored). I will add an issue to consider how to track the reference of all array elements based on registers information during runtime.

This makes sense. Thanks!

> From this conversation, I think maybe the most interested things of SeedLang clients are changes related named variables and elements of containers. So we can only focus on these changes, e.g. comparison of named variables and container elements, send notification of them, and ignore comparison of const of temporary variables. Is it reasonable?

I agree. No need to care about temporary variables.

The most significant request from SeedLang users would be monitoring the change of a variable or an element of a list/dict variable.

But, there are still complex things even if we simplify things like this. Here is one example:

```
a = [1, 2, 3]
b = a
b[2] = 10
```

How can the visualizer know the content of the list `a` has been changed when it receives the assignment notification about `b[2] = 10`?

Let's finish this PR. We can discuss those more complex problems later. I will also create a new issue to track the a/b problem as showed above.

wixette (approved):
LGTM

PR #194, opened 2022-05-14
Support Stop() from callback of notification events

### LINKED ISSUE #188: Support Stop() during the runtime of a SeedLang script
# Example case

The sorting example of SeedLang runs SeedPython scripts such as bubble_sort.py. It usually costs a rather long time (say, a few minutes) to run the code and play animations for each operation.

There will be two buttons on the UI of the sorting example: "Run" and "Stop". The user uses the "Run" button to start tunning a program. When the program is running and the visualization animations are being played, the user shall be able to click the "Stop" button to cancel the demonstration. - For doing this, the SeedLang engine may need to provide a Stop() interface.

# Possible solutions

1. SeedLang.Engine provides an interface `Stop()` for client code to call. Typically the client code will prefer calling this interface in the SingleStep event's On() callback.

2. Enable the SingleStep event to receive a return value from the client code, to indicate if the rest execution of the program should be continued or not.

codingpotato:
Fix #188

### DIFF
--- a/csharp/src/SeedLang/Engine.cs
+++ b/csharp/src/SeedLang/Engine.cs
@@ -19,6 +19,7 @@ using System.IO;
 using SeedLang.Ast;
 using SeedLang.Common;
 using SeedLang.Interpreter;
+using SeedLang.Runtime;
 using SeedLang.X;
 
 namespace SeedLang {
@@ -29,6 +30,9 @@ namespace SeedLang {
     // The semantic tokens of the source code. It falls back to syntax tokens when there are parsing
     // or compiling errors.
     public IReadOnlyList<TokenInfo> SemanticTokens => _semanticTokens;
+
+    // The state of SeedLang VM.
+    public VMState State => _vm.State;
 
     private readonly SeedXLanguage _language;
     private readonly RunMode _runMode;
@@ -97,21 +101,34 @@ namespace SeedLang {
 
     // Dumps the AST tree of the source code.
     public string DumpAst() {
-      Debug.Assert(!(_astTree is null));
+      Debug.Assert(!(_astTree is null), "Parse and compile source code before dumping AST tree.");
       return _astTree.ToString();
     }
 
     // Disassembles the compiled bytecode of the source code.
     public string Disassemble() {
-      Debug.Assert(!(_func is null));
+      Debug.Assert(!(_func is null),
+                   "Parse and compile source code before Disassembling bytecode.");
       return new Disassembler(_func).ToString();
     }
 
     // Runs the compiled bytecode of the source code.
     public bool Run(DiagnosticCollection collection = null) {
-      Debug.Assert(!(_func is null));
+      Debug.Assert(!(_func is null), "Parse and compile source code before running.");
       try {
         _vm.Run(_func);
+        return true;
+      } catch (DiagnosticException exception) {
+        collection?.Report(exception.Diagnostic);
+        return false;
+      }
+    }
+
+    // Continues execution of current program. The execution can be stopped from the callback
+    // function of visualization events.
+    public bool Continue(DiagnosticCollection collection = null) {
+      try {
+        _vm.Continue();
         return true;
       } catch (DiagnosticException exception) {
         collection?.Report(exception.Diagnostic);
--- a/csharp/src/SeedLang/Engine.cs
+++ b/csharp/src/SeedLang/Engine.cs
@@ -14,7 +14,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using SeedLang.Ast;
 using SeedLang.Common;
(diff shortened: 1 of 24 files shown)


csharp/src/SeedLang/Engine.cs
```diff
@@ -97,19 +101,20 @@ public class Engine {
 
     // Dumps the AST tree of the source code.
     public string DumpAst() {
-      Debug.Assert(!(_astTree is null));
+      Debug.Assert(!(_astTree is null), "Parse and compile source code before dumping AST tree.");
```
wixette, line 104: Given `DumpAst` is a public interface exposed to SeedLang users, it's better to return the error message (via Diagnostics) directly, instead of asserting a prediction.

csharp/src/SeedLang/Engine.cs
```diff
@@ -97,19 +101,20 @@ public class Engine {
...
       return _astTree.ToString();
     }
 
     // Disassembles the compiled bytecode of the source code.
     public string Disassemble() {
-      Debug.Assert(!(_func is null));
+      Debug.Assert(!(_func is null),
+                   "Parse and compile source code before Disassembling bytecode.");
```
wixette, line 111: Same as above.

csharp/src/SeedLang/Engine.cs
```diff
@@ -97,19 +101,20 @@ public class Engine {
...
       return new Disassembler(_func).ToString();
     }
 
     // Runs the compiled bytecode of the source code.
     public bool Run(DiagnosticCollection collection = null) {
-      Debug.Assert(!(_func is null));
+      Debug.Assert(!(_func is null), "Parse and compile source code before running.");
```
wixette, line 117: Same as above.

csharp/src/SeedLang/Engine.cs
```diff
@@ -119,6 +124,18 @@ public class Engine {
       }
     }
 
+    // Continues execution of current program. The execution can be stopped from the callback
+    // function of visualization events.
+    public bool Continue(DiagnosticCollection collection = null) {
```
wixette, line 129: Besides the `Continue` interface, a `Stop` or `Quit` interface is required too I think. If a visualizer pauses (via IVM.Stop()) the execution, it will have two choices: (1) continue the execution. (2) stop (quit from) the execution immediately. Or, we can also provide only `Continue` here and provide two interfaces in the IVM interface: `Pause` and `Quit`.

csharp/src/SeedLang/Visualization/IVM.cs
```diff
@@ -0,0 +1,22 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace SeedLang.Visualization {
+  // The interface of SeedLang VM. The methods of it can only be called during visualization
+  // notification.
+  public interface IVM {
+    // Stops execution.
+    void Stop();
```
wixette, line 20: 1. The actual meaning of this method is `Pause`, right? `Pause` is a better name than `Stop` here.

2. We also need a real `Stop` or `Quit` interface. This can be provided either in the IVM interface, or in the Engine class.

csharp/src/SeedLang/Runtime/VMState.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace SeedLang.Runtime {
+  // The state of SeedLang VM.
+  public enum VMState {
+    // The SeedLang VM is ready to execute a program.
+    Ready,
+    // The script is running.
+    Running,
+    // The execution is stopped temporarily.
+    Stopped,
```
wixette, line 23: `Paused` is a better name than `Stopped` here.

wixette (changes requested):

csharp/src/SeedLang/Engine.cs
codingpotato: Fixed.

csharp/src/SeedLang/Engine.cs
codingpotato: Fixed.

csharp/src/SeedLang/Engine.cs
codingpotato: Fixed.

csharp/src/SeedLang/Engine.cs
codingpotato: Add `Stop()` functions in `Engine` class and `IVM` interface.

csharp/src/SeedLang/Visualization/IVM.cs
codingpotato: Changed to `Pause` and added a real `Stop` interface.

csharp/src/SeedLang/Runtime/VMState.cs
codingpotato: Fixed.

wixette (approved):
LGTM

PR #202, opened 2022-05-29
Implement variable defined and deleted notification

codingpotato:
Implement variables information track in SeedVM and send variable defined and deleted notification to visualizers.

Will add variable subscription assignment tracking in the coming PRs.


wixette (approved):
LGTM

PR #203, opened 2022-06-03
Implement GetGlobals and GetLocals interfaces

codingpotato:
- Support get all global and local variables for IVM interface.
- Remove "global." prefix from the chained name of variables, because all variables have this prefix. It doesn't have significant meaning, and there is information in the visualization notification to indicate if the variable is global or local.


csharp/src/SeedLang/Engine.cs
```diff
@@ -18,21 +18,27 @@
...
 
 namespace SeedLang {
   // Engine is the facade class of the SeedLang core library. It provides interfaces of compiling
   // and executing SeedX programs, registering visualizers, and inspecting the execution status of
   // programs.
   public class Engine {
+    // The flag to indicate if variable tracking is enabled. It's enabled by calling
+    // "EnableVariableTracking", or registering variable related visualizers like "VariableDefined",
+    // "VariableDeleted" etc.
+    public bool VariableTrackingEnabled => _vm.VisualizerCenter.VariableTrackingEnabled;
```
wixette, line 31:
```suggestion
    public bool IsVariableTrackingEnabled => _vm.VisualizerCenter.VariableTrackingEnabled;
```

to be consistent with the following names such as "IsProgramCompiled"?

csharp/src/SeedLang/Engine.cs
```diff
@@ -64,6 +70,11 @@ public class Engine {
       _vm.VisualizerCenter.Unregister(visualizer);
     }
 
+    // Enables variable tracking.
+    public void EnableVariableTracking() {
```
wixette, line 74: How to turn off this switch? Or, consider a bool parameter?

```
public void EnableVariableTracking(bool enable)
```

Or, consider making the property writable?

wixette (changes requested):

csharp/src/SeedLang/Engine.cs
codingpotato: Done.

csharp/src/SeedLang/Engine.cs
codingpotato: Remove this function, and make the property writable.

wixette (approved):
LGTM

PR #204, opened 2022-06-06
Implement multiple-dimension array access notification

### LINKED ISSUE #189: Implement multiple-dimension elements access notification
This is very tricky for the 2-dimensional matrix access, such as:

matrix2x3 = [ [1, 2], [3, 4], [5, 6]]
matrix2x3[1][1] = 9
This is a common case in basic algorithms or data structure programs.

The current implementation only outputs the following event:

SubscriptAssign: [3, 9][1] = 9
What's the meaning of 3, 9 here?

It's hard for the visualizer to tell which variable is accessed.

Is it possible to divide the operation matrix2x3[1][1] = 9 into two phases:

phase 1: get the left value by accessing the memory position of matrix2x3[1], and assign the position to an unnamed variable.
phase 2: assign 9 to the unnamed variable.
And send out the info of both the two phases to the client side?

As for a way more complex case, like:

matrix[3][4][a][b][c][9][8] = 10
It's kind of a chained memory access operation. Can we send out the full memory access chain out so that the client understands where the value is actually put in?

This is really a complex problem. We might need more considerations and experiments here.

Also consider how a typical modern debugger/insepctor handles the code line:

matrix[3][4][a][b][c][9][8] = 10
Does a modern debugger/inspector know where in the matrix list an internal element is being changed?

codingpotato:
Fix #189 

Run following code in SeedLang.Shell to see the notifications.

```python
a = [1, 2, 3]
print(a)  # [1, 2, 3]

a[1] = [1, 2, 3]
print(a)  # [1, [1, 2, 3], 3]

a[1][1] = [1, 2, 3]
print(a)  # [1, [1, [1, 2, 3], 3], 3]

a[1][1][1] = 5
print(a)  # [1, [1, [1, 5, 3], 3], 3]
```


wixette (approved):
This is super! LGTM

PR #208, opened 2022-06-19
Include variable information into assignment, binary and comparison notification events

codingpotato:
- Change `Binary` and `Comparison` events to include variable information.
- Remove `SubscriptAssignment` events, and combine it into `Assignment` events.
- Remove `Boolean` and `Unary` events. It seems that these events are not useful now.

After this change, the `Compare VTag` is not needed any more. Please run following command to see the notification of `Assignment`, `Binary` and `Comparison`.

```shell
dotnet run --project csharp/src/SeedLang.Shell -- -v "*" -f examples/seedpython/scripts/sorting/bubble_sort.py
```

### DIFF
--- a/csharp/src/SeedLang/Interpreter/CompilerHelper.cs
+++ b/csharp/src/SeedLang/Interpreter/CompilerHelper.cs
@@ -203,16 +203,10 @@ namespace SeedLang.Interpreter {
 
     internal void EmitSubscriptAssignNotification(uint containerId, uint keyId, uint valueId,
                                                   TextRange range) {
+      // Check if there are assignment events registered in the visualizer center, because SeedVM
+      // will send assignment events for subscript assignments as well.
       if (!_suspendNotificationEmitting && _visualizerCenter.HasVisualizer<Event.Assignment>()) {
         var n = new Notification.SubscriptAssignment(containerId, keyId, valueId);
-        Emit(Opcode.VISNOTIFY, 0, Cache.IdOfNotification(n), range);
-      }
-    }
-
-    internal void EmitUnaryNotification(UnaryOperator op, uint valueId, uint resultId,
-                                        TextRange range) {
-      if (!_suspendNotificationEmitting && _visualizerCenter.HasVisualizer<Event.Unary>()) {
-        var n = new Notification.Unary(op, valueId, resultId);
         Emit(Opcode.VISNOTIFY, 0, Cache.IdOfNotification(n), range);
       }
     }
--- a/csharp/src/SeedLang/Interpreter/CompilerHelper.cs
+++ b/csharp/src/SeedLang/Interpreter/CompilerHelper.cs
@@ -84,8 +84,10 @@ namespace SeedLang.Interpreter {
       return _variableResolver.FindVariable(name);
     }
 
-    internal uint DefineTempVariable() {
-      return _variableResolver.DefineTempVariable();
+    internal uint DefineTempVariable(TextRange range) {
+      uint id = _variableResolver.DefineTempVariable();
+      EmitTempRegisterAllocatedNotification(id, range);
+      return id;
     }
 
     internal uint? GetRegisterOrConstantId(Expression expr) {
@@ -153,12 +155,8 @@ namespace SeedLang.Interpreter {
         var nId = Cache.IdOfNotification(new Notification.Function(name, funcRegister, argLength));
         Emit(Opcode.VISNOTIFY, (uint)Notification.Function.Status.Returned, nId, range);
       }
-      bool notifyVariableDeleted = !_suspendNotificationEmitting &&
-                                   !NativeFunctions.IsNativeFunc(name) &&
-                                   _visualizerCenter.IsVariableTrackingEnabled;
-      if (notifyVariableDeleted) {
-        var nId = Cache.IdOfNotification(new Notification.VariableDeleted(0));
-        Emit(Opcode.VISNOTIFY, 0, nId, range);
+      if (!NativeFunctions.IsNativeFunc(name)) {
+        EmitVariableDeletedNotification(0, range);
       }
     }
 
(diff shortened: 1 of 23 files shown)


csharp/src/SeedLang/Visualization/EventData.cs
```diff
@@ -0,0 +1,101 @@
...
+    Global,
+    Local,
+    Upvalue,
+  }
+
+  // The class that is used as operands of binary and comparison events. It could be a variable,
+  // an element of containers or a temporary value. The Variable field is null if it's a temporary
+  // value.
+  public class Operand {
+    public bool IsVariable => !(Variable is null);
```
wixette, line 30: The structure of `Operand` and `Variable` is a bit confusing:

First, the current `Operand` is classified to:
- Is a Variable
- Is NOT a Variable (that is, a temporary value)

Second, the current `Variable` is classified to:
- Is an Element of a Container
- Is NOT an Element of a Container (that is, a normal variable)

Theoretically "an element of a container" is not always treated as a variable, right?

Can we define an Enum with three categories to classify an operand?
- Variable operand
- ElementOfContainer operand
- TemporaryValue operand

wixette (changes requested):

csharp/src/SeedLang/Interpreter/CompilerHelper.cs
```diff
@@ -195,21 +203,14 @@ internal class CompilerHelper {
 
     internal void EmitSubscriptAssignNotification(uint containerId, uint keyId, uint valueId,
                                                   TextRange range) {
-      if (!_suspendNotificationEmitting &&
-          _visualizerCenter.HasVisualizer<Event.SubscriptAssignment>()) {
+      // Check if there are assignment events registered in the visualizer center, because SeedVM
```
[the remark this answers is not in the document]
codingpotato: Done.

csharp/src/SeedLang/Visualization/EventData.cs
codingpotato: Fixed. Please see my comments seperately.

codingpotato:
Fixed the review comments and added another big change to fix bugs in variable tracking.

1. Change the event data of visualization notification to LValue and RValue, so that the right value of assignment can also be represented as `Variable`, `Element of Containers` and `Temporary Value`.
2. Added temporary register allocated notification to make variable tracking correct. It will add more instructions when the client would like to get variable related notifications. It makes the program ~5 times slower than normal case without variable tracking enabled. The performance is acceptable now.
3. The variable tracking logic is quite complex. So there are some minor bugs now. I will check and fix them later.
4. Please run following command to see the result of this PR.
```shell
dotnet run --project csharp/src/SeedLang.Shell -- -v "Assignment,Comparison" -f examples/seedpython/scripts/sorting/bubble_sort.py
```

With this PR, the comparison notification events look like
```text
6             if a[j] > a[j + 1]:
Comparison: a:Global[0] 0 > a:Global[1] 1 = False
```
And assignment notification events look like
```text
8                 a[j], a[j + 1] = a[j + 1], a[j]
Assign: a:Global[5] = a:Global[6] 1
8                 a[j], a[j + 1] = a[j + 1], a[j]
Assign: a:Global[6] = a:Global[5] 5
```
Please have a review again.

csharp/src/SeedLang/Visualization/EventData.cs
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  public enum RValueType {
+    ElementOfContainer,
+    TemporaryValue,
+    Variable,
+  }
+
+  // The LValue class. It could be a variable or an element of containers. The Keys field is null if
+  // it's a variable.
```
wixette, line 39: For a better interface for this class, it's also good to provide properties to let the client code easily determine the type. E.g.,

```
public bool IsVariable => key is null;
public bool IsElement => !IsVariable;
```

csharp/src/SeedLang/Visualization/EventData.cs
```diff
@@ -0,0 +1,144 @@
...
+        LValueType.ElementOfContainer => $"{Variable}[{string.Join("][", Keys)}]",
+        LValueType.Variable => $"{Variable}",
+        _ => throw new NotImplementedException($"Unsupported LValue type: {Type}"),
+      };
+    }
+  }
+
+  // The RValue class. It could be a variable, an element of containers or a temporary value. The
+  // Keys field is null if it's a variable. The Variable and Keys field is null if it's a temporary
+  // value.
```
wixette, line 68: As above:

```
bool IsVariable => keys is null;
bool IsTemporary => Variable is null && Keys is null;
```

csharp/src/SeedLang/Visualization/EventData.cs
```diff
@@ -0,0 +1,144 @@
...
+      return Type switch {
+        LValueType.ElementOfContainer => $"{Variable}[{string.Join("][", Keys)}]",
+        LValueType.Variable => $"{Variable}",
+        _ => throw new NotImplementedException($"Unsupported LValue type: {Type}"),
+      };
+    }
+  }
+
+  // The RValue class. It could be a variable, an element of containers or a temporary value. The
+  // Keys field is null if it's a variable. The Variable and Keys field is null if it's a temporary
```
wixette, line 67:
```suggestion
  // Keys field is null if it's a variable. The Variable and Keys field are null if it's a temporary
```

csharp/src/SeedLang/Visualization/EventData.cs
codingpotato: Done.

csharp/src/SeedLang/Visualization/EventData.cs
codingpotato: Done.

csharp/src/SeedLang/Visualization/EventData.cs
codingpotato: Done.

wixette (approved):
LGTM

PR #210, opened 2022-06-25
Fix extra tab in the last line parsing issue

### LINKED ISSUE #209: Indention but no EOF in the last line causes compilation errors
```
      var engine = new Engine(SeedXLanguage.SeedPython, RunMode.Script);
      string code = "a=0\nfor i in range(10):\n\ta = a + i\n\t";
      var collection = new DiagnosticCollection();
      if (!engine.Compile(code, "", collection)) {
        Console.WriteLine(collection.Diagnostics[0].ToString());
      } else {
        Console.WriteLine("OK");
      }
```

The above code results in SeedLang compilation error:

```
20220620002906110 Fatal (SeedLang.X) <> [Ln 4, Col 1 - Ln 4, Col 0] 29: SyntaxErrorMissingToken '<EOF>' DEDENT
```

While in Python3, the same code piece can be executed with no error:

```
Python 3.9.13 (main, May 24 2022, 21:13:51) 
[Clang 13.1.6 (clang-1316.0.21.2)] on darwin
>>> code = "a=0\nfor i in range(10):\n\ta = a + i\n\t"
>>> exec(code)
>>> a
45
```

codingpotato:
Fix #209

### DIFF
--- a/csharp/src/SeedLang/X/SeedPythonDentLexer.cs
+++ b/csharp/src/SeedLang/X/SeedPythonDentLexer.cs
@@ -104,7 +104,7 @@ namespace SeedLang.X {
         spacesStart++;
       }
       int next = InputStream.LA(1);
-      if (next == SeedPythonParser.Eof) {
+      if (spacesStart == Text.Length && next == SeedPythonParser.Eof) {
         _addTrailingNewline = false;
       }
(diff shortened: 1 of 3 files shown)


wixette (approved):
LGTM

PR #211, opened 2022-06-26
Compile non-comparison and non-boolean expressions in the boolean expressions

### LINKED ISSUE #159: Boolean expression with comparison and boolean constant crash issue
SeedLang.Shell crashes when running following code:

```python
if 1 < 2 and True:
  print(1)
```

codingpotato:
Fix #159 

Support following code now:
```python
def is_even(n):
  return n % 2 == 0


for n in range(10):
  if n < 5 and is_even(n):
    print(n)
```


wixette (approved):
LGTM

PR #212, opened 2022-06-26
Bug fix and performance optimization for variable tracking

codingpotato:
There isn't new functions implemented in this PR.
- One critical bug fix for variable deleted notification logic.
- Replace temp register allocated notification with temp register freed notification to improve the performance.


wixette (approved):
LGTM

Issue #213, opened 2022-06-27
Implement python runtime library


codingpotato:
- math library

codingpotato:
Fixed by #218

PR #218, opened 2022-07-17
Implement functions of the math library

codingpotato:
Implement #213 

Please run math_lib.py for example code of the math library.

```shell
dotnet run --project csharp/src/SeedLang.Shell -- -f examples/seedpython/scripts/math_lib.py
```

Some builtin functions are added as well, like abs, all, any, dir, sum. For example:
```python
a = [True, False, True]
all(a)
any(a)

a = [1, 2, 3]
sum(a)

dir()

import math
dir(math)
```


csharp/src/SeedLang/Interpreter/VM.cs
```diff
@@ -32,45 +32,51 @@ private enum State {
...
+    private Module _module;
     private CallStack _callStack;
     private Chunk _chunk;
     private int _pc;
 
-    // The hash table to store defined global variable names.
-    private HashSet<string> _globals;
+    // Returns the names in the module namespace, if the given value is a module object. Otherwise
+    // returns an empty list.
+    internal static VMValue Dir(VMValue value) {
```
wixette, line 54: "Dir" is a bit confusing with the concept "directory" as in a filesystem.

Is "ModuleDir" better here?

csharp/src/SeedLang/Interpreter/VM.cs
```diff
@@ -32,45 +32,51 @@ private enum State {
...
       _visualizerCenter = visualizerCenter;
     }
 
     internal void RedirectStdout(TextWriter stdout) {
-      _sys.Stdout = stdout;
+      Stdout = stdout;
+    }
+
+    // Returns the names in current module namespace.
+    internal VMValue Dir() {
```
wixette, line 70: Same as above.

csharp/src/SeedLang/Runtime/BuiltinsDefinition.cs
```diff
@@ -0,0 +1,213 @@
...
+    public const string Dir = "dir";
+    public const string Len = "len";
+    public const string List = "list";
+    public const string Print = "print";
+    public const string Range = "range";
+    public const string Round = "round";
+    public const string Slice = "slice";
+    public const string Sum = "sum";
+
+    public static Dictionary<string, VMValue> Variables = new Dictionary<string, VMValue> {
```
wixette, line 39: I think this definition dictionary initialization process can be simplified by C#'s Attribute classes.

For example, considering defining module method function like below:

```
[ModuleFunc(name = PrintVal)]
private static VMValue PrintValFunc(ValueSpan args, INativeContext context) {
  ...
}
```

With the above attribute-decorated functions, the initializer can access the attribute info of every function then create the dictionary accordingly.

Cons of the above solution: it requires C# reflection to read attribute info. I am not sure if Unity's managed code stripping https://docs.unity3d.com/Manual/ManagedCodeStripping.html will remove the reflection logic somehow.

csharp/src/SeedLang/Interpreter/VM.cs
codingpotato: Changed to "ModuleDir".

csharp/src/SeedLang/Runtime/BuiltinsDefinition.cs
codingpotato: Re-implement by reflection without attribute. We can change the implement later if Unity removes these logic after testing in Unity 3D environment.

wixette: OK. This solution looks good for now.

wixette (approved):
LGTM

PR #221, opened 2022-07-26
Change field name from Value to RValue for assignment events

### LINKED ISSUE #217: The Value.Value property of Assignment Event
```csharp
public void On(Event.Assignment e, IVM vm) {
    if (... && e.Value.IsTemporary && e.Value.Value.IsList) {
  ...
```

Think about how to change the weird Value.Value.

codingpotato:
Fix #217 

The client could access the value of assignment events by:

```csharp
double number = assignment.RValue.Value.AsNumber();
```


wixette (approved):
LGTM

PR #232, opened 2022-08-14
Localize SeedLang messages

wixette:
And, let SeedLang.Shell support a new option to set the locale of the SeedLang engine.


codingpotato (approved):
LGTM

